PR #564, opened 2019-04-25
Dropdown different label and value

### LINKED ISSUE #552: Azure Function runtime - node, correct? - 0.0.19112-4
image

is this correct?  shouldn't it be Node.JS or something

dandua98:
This PR fixes #552. Dropdowns on function modal can have different labels and values now, allowing us to use a different internal and displayed name for dropdowns.

_**Testing Notes:**_

- Launch the extension in VSCode
- Navigate to Services page and add Azure Functions. You'll see that runtime stack shows *JavaScript* instead of *node* now. The internal name used is still *node* since Azure ARM uses *node* internally.
- Make sure everything else is still working regularly


crutkas:
Any reason why we are afraid to check this in?

sibille:
Don't think so, will test and integrate.

dandua98:
Just FYI, we had this tested and ready to merge but since it was on the last day (of the internship) and we were using dev for build, we decided to leave this open and not merge it in. I think it should be good to merge if it's still functional (I believe there have been a lot of new changes since)

sibille (approved):

Issue #570, opened 2019-04-25
Bootstrap list template should add an item on Enter Key press


dandua98:
![image](https://user-images.githubusercontent.com/24802378/56774713-c035d280-6778-11e9-88e1-c534663dcd77.png)

Pressing enter on a list item should add the item to list, this would be more intuitive.

crutkas:
angular should be updated as well as react

crutkas:
i think we also need to have maybe documented page behaviors so when new frameworks come online / something is adjusted, it is documented it should act the same across the board.

crutkas:
This is another great chance to maybe add in a web ui test :)

streetcoder123:
This issue has been fixed

PR #607, opened 2019-05-23
Allow list entry to be submitted on enter key press - React Application

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/26876
Github Issue: #570 

**Changes**
- React List Template should be able to submit a list item both by pressing the submit button and on enter key press.
- This is achieved using the onSubmit command in the form tag rather than using it in the submit button.


SahilTara:
Hey,
Just wondering if you could also tackle angular's too in a seperate PR (or this one since related problem with same solution) since I suspect it will be a problem. https://github.com/microsoft/WebTemplateStudio/blob/dev/templates/Web/Pages/Angular.List/src/app/components/AngularList/ListForm/list-form.component.html

Syntax slightly different but shouldn't be too bad.

Edit: I would look into ngSubmit.

src/client/src/translations/whitelist_en.json
```diff
@@ -68,7 +68,6 @@
   "cosmosResourceModule.saveChanges",
   "cosmosResourceModule.createCosmosRes",
   "cosmosResourceModule.internalName",
-  "footer.license",
```
jsondoo, line 71: why was this line removed?

SahilTara: I'm guessing this was due to running build since footer.license isnt used anywhere else but I may be wrong. This shouldn't be removed technically, but it reveals an underlying issue https://github.com/microsoft/WebTemplateStudio/blob/c3eeff139a294756156842a7cf394d254d500f34/src/client/src/containers/Footer/index.tsx
Looking at this file it seems that footer.license isnt used anywhere, it should be doing similar to what the aria label is doing with intl.

streetcoder123:
> Hey,
> Just wondering if you could also tackle angular's too in a seperate PR (or this one since related problem with same solution) since I suspect it will be a problem. https://github.com/microsoft/WebTemplateStudio/blob/dev/templates/Web/Pages/Angular.List/src/app/components/AngularList/ListForm/list-form.component.html
> Syntax slightly different but shouldn't be too bad.
> Edit: I would look into ngSubmit.

Yes, I am creating a separate PR for angular. It should also be there soon.

crutkas:
what is the actual issue on Github this relates to.

jsondoo:
#570

src/client/src/translations/whitelist_en.json
crutkas: @njindal239 why was the line removed

crutkas:
why did the pipeline fail?  @sibille the logs are very spare here

crutkas:
@njindal239  also looks like there is a branch conflict now

src/client/src/translations/whitelist_en.json
streetcoder123: This change comes in when I run ./build on the project. I am not sure whether it needs to be there or not but since it came out of building the project, I committed it with my changes.

sibille:
@crutkas, seems the build failed during initialization, went fine when I hit "Rebuild".

SahilTara (dismissed):
Everything seems to work as intended for the list. Code looks sound. The line that was removed due to ./build needs an issue against it since we are not using intl properly for that one parameter. #614

piotrmark (approved):

templates/Web/Pages/React.List/src/components/ReactList/ListForm.jsx
```diff
@@ -12,9 +12,9 @@ export default function ListForm(props) {
         placeholder="Add text here..."
         aria-label="Add text here..."
       />
-      <button className="btn btn-primary ml-2" onClick={onAddListItem}>
+      <button className="btn btn-primary ml-2">
```
piotrmark: I think it's better to explicitly define type="submit" on the button. It's default for most of the browsers, but per W3Schools: "Always specify the type attribute for the <button> element. Different browsers may use different default types for the <button> element."
https://www.w3schools.com/tags/att_button_type.asp

piotrmark (changes requested):

templates/Web/Pages/React.List/src/components/ReactList/ListForm.jsx
streetcoder123: That's a good catch. I will update that. Thanks!

jsondoo (approved):

piotrmark (approved):

Tanya0609:
I did review. It works so approve this change

Tanya0609 (approved):

templates/Web/Pages/React.List/src/components/ReactList/ListForm.jsx
SahilTara: Really good catch!

Issue #589, opened 2019-05-16
Ensure at least one page is generated


sibille:
The wizard should not allow to delete all pages, at least one page should be generated.

PR #618, opened 2019-05-27
GitHub issue #589 ensure at least one page is generated

Tanya0609:


SahilTara:
Can you please attach the issue related to this PR?

sibille (commented):
Works fine, but I get a warning on the code saying "Avoid typeof x === 'undefined' comparisons." Think this should be "(totalPageNumber !== undefined? totalPageNumber > 1 : true)" instead. 
Also, can this ever be undefined? or would "(totalPageNumber > 1)" be sufficient?

crutkas:
@Tanya0609 what issue is this for?  Please assign the issue to yourself and xref it to the PR (use # and the issue number here)

Tanya0609:
The GitHub Issue the PR is related to:
https://github.com/microsoft/WebTemplateStudio/issues/589

Change:
When there is only one page is selected, the close (x) button won't show up anymore in the side bar.

src/client/src/components/DraggableSidebarItem/index.tsx
```diff
@@ -139,12 +141,13 @@ const DraggableSidebarItem = ({
             </div>
           )}
         </div>
-        <CloseSVG
+        {(typeof totalPageNumber !== 'undefined'? totalPageNumber>1 : true) && <CloseSVG
```
SahilTara: Generally it is better to do totalPageNumber !== undefined as @sibille mentioned. 
```suggestion
        {(totalPageNumber !== undefined ? totalPageNumber > 1 : true) && <CloseSVG
```

src/client/src/components/SortableSelectionList/index.tsx
```diff
@@ -105,6 +108,7 @@ const SortableList = SortableContainer(
                   page={page}
                   handleInputChange={handleInputChange}
                   handleCloseClick={handleCloseClick}
+                  totalPageNumber = {pages.length}
```
SahilTara: Very small question will mark this as resolved since it isn't a required changed but wondering why you check the pages.length inside the map function every time rather than just create something like ``` const pageLength =  pages.length ``` then use that inside the map? May just be a design decision and that's totally fine.

SahilTara (changes requested):

src/client/src/components/SortableSelectionList/index.tsx
SahilTara: Turns out I cannot mark it as resolved since I cannot request changes 😞. All good though this keeps visibility to the question.

Tanya0609: Good point, I will update my code

src/client/src/components/DraggableSidebarItem/index.tsx
Tanya0609: updated my change

SahilTara (dismissed):

jsondoo:
@sibille is there a specific linter that we should be using?

SahilTara:
@jsondoo use the tslint extension. I forget which one though since one worked and the other was buggy.

jsondoo:
image
image

@SahilTara  Seems there's two main ones, the deprecated one has higher rating, should we go with the Microsoft one?

SahilTara:
@jsondoo i would try both isolated, one of them consistently showed warnings. I'm not at home right now though so I can't tell you the exact one at the moment.

src/client/src/components/SortableSelectionList/index.tsx
```diff
@@ -33,6 +35,7 @@ const SortableSidebarItem = SortableElement(
         handleInputChange={handleInputChange}
         handleCloseClick={handleCloseClick}
         idx={idx + 1}
+        totalPageCount = {totalPageCount}
```
jsondoo: for consistency, should we have `totalPageCount={totalPageCount}` (removing the spaces)

src/client/src/components/SortableSelectionList/index.tsx
```diff
@@ -105,6 +109,7 @@ const SortableList = SortableContainer(
                   page={page}
                   handleInputChange={handleInputChange}
                   handleCloseClick={handleCloseClick}
+                  totalPageCount = {totalPageCount}
```
jsondoo: same thing here

jsondoo (commented):
good work 👍

streetcoder123 (dismissed):
It looks good and works as Expected 👍 
Incorporate those small consistency things pointed by @jsondoo and this PR seems good to go

SahilTara (dismissed):

crutkas:
can we get a couch / FTE to do a review?

src/client/src/components/DraggableSidebarItem/index.tsx
```diff
@@ -139,12 +141,13 @@ const DraggableSidebarItem = ({
             </div>
           )}
         </div>
-        <CloseSVG
+        {(totalPageCount !== undefined ? totalPageCount > 1 : true) && <CloseSVG
```
piotrmark: NIT: Style - I think it looks better (and is consistent with other tags) to start HTML tag in new line.
BUT: I believe that code formatting is not developer's job - it's computer's job. @SahilTara, Did you guys use an autoformatting tool? I remember hearing the word "Prettier" in one of onboarding videos :)
If the answer is yes, then everyone should configure "format on save" and don't think about style any more :)

piotrmark (dismissed):

mrlacey:
> can we get a couch / FTE to do a review?

that's a fun typo 😉 🛋

src/client/src/components/DraggableSidebarItem/index.tsx
SahilTara: Yeah we used prettier a lot! I too recommend everyone gets prettier. Turn format on save on or else there will be giant diffs between people who have format on save on and those who don't (we experienced such). https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode

mrlacey: If you're going to change the formatting of code, it's beneficial to make the reformatting changes as a separate commit so changes in functionality are clear and separate.
Also it's important that everyone is using the same layout/formatting/linting rules as everyone else. Having consistency across the code base makes reviews simpler and helps when maintaining code.

If it is decided that a project should use a particular tool then then should be specified as a recommended extension in the workspace. (see https://code.visualstudio.com/updates/v1_6#_workspace-extension-recommendations)
Any configuration files for extensions should also be checked into the repo so that everyone uses the same tools with the same settings.

jsondoo (dismissed):

Tanya0609:
I did a rebase to newest dev, so that it won't cause conflicts when merging later

src/client/src/components/DraggableSidebarItem/index.tsx
Tanya0609: fixed as suggested. The formatting change is in another commit

streetcoder123 (approved):

piotrmark (approved):

jsondoo (approved):

Issue #599, opened 2019-05-19
AngularJS should rename to just Angular


doggy8088:
I've just seen "**AngularJS**" in the WTS preview mode. The name **AngularJS** should correct to **Angular** for better understanding.

crutkas:
Agreed. Their docs reference angular. https://github.com/angular/angular

sibille:
@jsondoo, can we change the internal name to Angular for consistensy too? You'll have to change it on the fronendframework.json, and also on the projects and page templates + compositions too

jsondoo:
@sibille yes I will look into it

PR #628, opened 2019-05-27
Change internal name to Angular from AngularJS

jsondoo:
Change internal name to Angular from AngularJS
Resolving follow-up on #599 @sibille 

To test you should run ./build and also repackage templates


SahilTara (commented):
Will test this out later in the day, but it seems good at first glance. Will give an approval once I've ensured that it works (it most likely does but still have to test!).

jsondoo:
Thanks @SahilTara, really appreciate the help

SahilTara:
Yup just tested everything seems to be good 😄

SahilTara (approved):

streetcoder123 (approved):
This works Good 👍

PR #629, opened 2019-05-27
Angular - List item should be submitted using Enter Key Press

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/27032
Github Issue: #570 

**Changes**
- The List Template generated for an Angular project should allow a list item to be submitted using the Enter key press along with submit button click.
- This is achieved by having a more general form submission event rather than a button click event which would handle all cases for form submission.
- Also, button type changed from 'button' to 'submit' so that the general form submission event could also cover the click on the submit button.


streetcoder123:
I am closing this PR as it did not follow the right branch name convention. Another PR opened to address the same issue: #630

PR #630, opened 2019-05-27
Angular - List item should be submitted using Enter Key Press

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/27032
Github Issue: #570
Changes
•	The List Template generated for an Angular project should allow a list item to be submitted using the Enter key press along with submit button click.
•	This is achieved by having a more general form submission event rather than a button click event which would handle all cases for form submission.
•	Also, button type changed from 'button' to 'submit' so that the general form submission event could also cover the click on the submit button.


templates/Web/Pages/Angular.List/src/app/components/AngularList/ListForm/list-form.component.html
```diff
@@ -1,16 +1,16 @@
-﻿<div class="input-group my-3">
+﻿<form (submit)="onAddListItem()" class="input-group my-3">
     <input
       type="text"
+      [ngModelOptions]="{standalone: true}"
```
SahilTara, line 4: Seems good in this case ngModelOptions is an okay choice. We could have went with defining the name property too.

SahilTara (approved):

annielovesjs (approved):
looks good! Tested the extension with these changes and it works

templates/Web/Pages/Angular.List/src/app/components/AngularList/ListForm/list-form.component.html
streetcoder123: I added `name` attribute to the input tag. Though it worked, it gave me some weird UI things. When I am typing, ti starts to give me suggestions from those list items that are already on the board, which was annoying. This does not seem to be the case with `noModelOptions`.

SahilTara: Yeah that's one of the differences.

markAtMicrosoft (approved):

templates/Web/Pages/Angular.List/src/app/components/AngularList/ListForm/list-form.component.html
xWiiLLz: > I added `name` attribute to the input tag. Though it worked, it gave me some weird UI things. When I am typing, ti starts to give me suggestions from those list items that are already on the board, which was annoying. This does not seem to be the case with `noModelOptions`.

I think what you're referring to is the default behavior of web browsers in regards to already filled form elements. I don't think bypassing this default behavior with `[ngModelOptions]` is a good choice, since it doesn't comply with [accessibility standards](https://www.w3.org/WAI/WCAG21/quickref/?versions=2.0#qr-ensure-compat-rsv). IMO you should bring back the `name` attribute.

SahilTara: Hey @xWiiLLz do you mind making an issue for this? Thank you so much for calling this out.

PR #633, opened 2019-05-28
Fixed Error Handling Bug for Specific "Untracked" Commands

trevorNgo:
Wrap function call to catch errors, leverage telemetry module _callWithTelemetryAndCatchHandleErrors_ to log in AppInsights and launch report an issue experience

Lint to follow standard Prettier

PR #634, opened 2019-05-28
Singleton design to controller

trevorNgo:
Removes null check in extension.ts and enforces singleton pattern for controller;
To follow #633


SahilTara:
this is how it should have been done instead of that "Abstract" thing that we had at one point.

src/extension/src/controller.ts
```diff
@@ -66,8 +71,25 @@ export class Controller {
...
+   * @param message The payload received from the wizard client. Message payload must include field 'module'
+   * @returns Singleton Controller type
+   */
+  public static getInstance(
+    context: vscode.ExtensionContext,
+    extensionStartTime: number
+  ) {
+    if (this._instance) {
+      this._instance.showReactPanel();
+    }
```
jcoc611-microsoft, line 85: NIT: would it be more clear to have `else { this._instance = ... }` here instead of as an or statement below?

trevorNgo: I agree it is es6 magic... your call

jcoc611-microsoft: Yeah...you can leave it like this, it is not a big dieal

annielovesjs (dismissed):

SahilTara:
Hey @annielovesjs wondering if you accidentally approved this PR due to the issue you pointes out on the PR this PR depends on (#633)

annielovesjs:
@SahilTara Hey! I was actually testing both of them individually (and tested this one twice) and this PR doesn't output the error in PR #633.

Tanya0609:
@trevorNgo How can I check if this PR? and what are the expected result? Or I just use the wizard to generate a template to see if it breaks anything?

src/extension/src/controller.ts
jsondoo: I think having an else is a lot more clearer because it is a little bit confusing

jsondoo: we will do it in a separate PR

jasmingoh (approved):

jsondoo (approved):

Issue #622, opened 2019-05-27
Remove the "3." from the Frameworks page to simplify title


xtinah-w:
Change "3. Select a front-end framework" to "Select a front-end framework" and similarly for the back-end framework tag.

jasmingoh:
https://github.com/microsoft/WebTemplateStudio/pull/635

PR #635, opened 2019-05-28
Removing number from selection pages

jasmingoh:
Resolving issue 662 (https://github.com/microsoft/WebTemplateStudio/issues/622)

Talked to Christina (designer on the team) about having numbers on the selection pages. We looked at Windows Template Studio as reference, and found out that they do not have numbers on the selection pages. Hence, we decided to make this change.

Here's the before and after for the Frameworks page:
BEFORE

AFTER

Please note: GitHub issue 662 only mentioned the Frameworks page, but this PR includes changes to other pages as well (i.e. the Project Type page, Pages page, Services page and Summary page).


streetcoder123 (dismissed):
Looks Good 👍

piotrmark (changes requested):
Please, run:
npm run manage:translations from src/client to update translations files.
This needs to be done after each string change

Edit: oops, I missed the en.json file. Looks good :)

piotrmark (dismissed):

src/client/src/containers/SelectFrontEndFramework/index.tsx
```diff
@@ -50,7 +50,7 @@ type Props = IDispatchProps & ISelectFrontEndFrameworkProps & InjectedIntlProps;
 const messages = defineMessages({
   selectFrontendFramework: {
     id: "selectFrontendFramework.selectFrontendFramework",
-    defaultMessage: "3. Select a front-end framework."
+    defaultMessage: "Select a front-end framework."
```
SahilTara, line 53: Notice how you modify "id: "selectFrontendFramework.selectFrontendFramework"," -> Read next comment

src/client/src/translations/en.json
```diff
@@ -115,12 +115,12 @@
   "projectName.projectName": "Project Name",
   "projectName.projectNameTitle": "Project Name:",
   "projectNameError.invalidRegex": "Name may only contain letters, numbers, dashes or underscores",
-  "review.frameworks": "3. Frameworks",
-  "review.pages": "4. Pages",
-  "review.projectType": "2. Project Type",
-  "review.reviewAndGenerate": "6. Your project summary.",
-  "review.services": "5. Services",
-  "review.welcome": "1. Welcome",
+  "review.frameworks": "Frameworks",
```
SahilTara, line 118: You modify review.frameworks.
image
The actual thing is untouched though :( am I missing something here?

SahilTara (changes requested):

src/client/src/translations/en.json
SahilTara: Weird i tried doing the update command but its not detecting it for some reason.

SahilTara: Turns out its because react intl doesn't detect string changes by default. deleting the file (manual)"en.json", then and adding it back by running ```npm run manage:translations``` will work.

jasmingoh: I see.. I saw @piotrmark 's comment too and ran the command. Seems like there's some changes to whitelist_en.json as well. Is that expected?

SahilTara: Yeah both are fine, in fact delete both and let the thing regenerate itself. They are both automated things so regeneration won't do any damage!

SahilTara: When we actually do full translations of such (after clint sends it off for translation that is when you want to be more careful with the translation jsons, but right now they're mutable).

jasmingoh: Also, I followed the steps you (@SahilTara) mentioned, and it seems like other things not related to this issue are changed. Should they be added to this PR as well? 

Here's an example:
image

SahilTara: Honestly for this auto generated file I would say yes its okay, because its 100% auto-generated and you have no real control over it. This also exposes that this was due for an update for a while. @piotrmark can correct me if i'm wrong here. Worst case I would try just deleting the suspect lines from the whitelist and en.json, and seeing if they regenerate if you're worried about it.

SahilTara: Actually you know what yeah just do ```git reset --hard HEAD``` (to revert changes, make sure you have nothing important uncommitted) then just remove the lines you modified from the en.json and see if they regenerate. That seems like the better option to keep things scoped and then make a seperate pr for updating the rest of the json? Like honestly your call here. What do you prefer since it doesn't make that big of a difference for an auto generated file (when its the file that would have been modified anyways in some way shape or form in the pr)

annielovesjs (approved):
looks good! All the changes are evident and everything runs perfectly

src/client/src/translations/en.json
jasmingoh: Thanks Sahil! I manually deleted the en.json file and re-ran `npm run manage:translations`, seems like the lines relevant to this PR are generated properly. 
Your suggestion sounds good; I'll create a separated issue to update the rest of the en.json file then. (Update: here's the [new GitHub issue](https://github.com/microsoft/WebTemplateStudio/issues/636))

piotrmark (approved):

SahilTara (approved):

Issue #632, opened 2019-05-28
We need to change our local development DLLs to debug mode.


SahilTara:
Currently developing templates locally is relatively hard. We need to do packaging locally and this takes  time from all people developing. This also makes testing features a bit harder and causes many merge conflicts. 

I suggest we change the DLLs to debug locally. Changing to debug will allow us to use the raw templates folder without having to:
1) Keep track of template versions
2) Spending time packaging
3) Fixing unnecessary and completely avoidable merge conflicts

@crutkas wondering what your opinion on this is. I have a hunch that this will speed up development on templates significantly. This won't affect release status in our VSIX since it rebuilds CORETS in the build pipeline with the release mode flag. Will only affect local development process.

This will also allow people to work on templates from any OS since currently the packaging tool relies on .NET Framework which is windows only, so testing locally becomes unnecessarily tedious for people not using windows. 

## Will need the following changes:
- [x] change macOS dlls to debug
- [ ] change linux dlls to debug
- [x] change windows dlls to debug
- [x] Add an environment check for debug mode or production, and choose the sync path accordingly
~add src/extension/src/api/templates to the .gitignore
Possibly will require the following change:
add copying the templates folder to the build script or create a ./build for local development, if it is used by the build pipeline. Or we could just make this copying the manual process.~

sibille:
Agree, local development should be against core dlls built in debug mode, so you can sync templates with the localtemplatesource, that accesses templates folder using a junction pointing directly to the templates folder at the project root. I don't think this requires adding src/extension/src/api/templates to the .gitignore, but just core dlls in debug mode and passing the correct path on sync (we do it this way in WinTS).

SahilTara:
Oh so you mean just changing the path in the build pipeline instead of doing the copying?

Edit: The reason I mentioned the git ignore was due to the redundant templates folder that would result from my solution. If we just change the path in the pipeline for our release this will also work without having to do the git ignore.

sibille:
In WinTS we have a IF debug statement on the code, if your in debug you use the path to the template folder, if you're in release you use the path to the mstx. 
Is there sth like this in React?

SahilTara:
Ah yes, there is! (In node and vscode). Much simpler solution thanks for the input will modify issue to incorporate this. Will need to test it out but I have a hunch on the solution.

jsondoo:
@SahilTara So what would the local development process look like after these changes?

SahilTara:
[...]

PR #639, opened 2019-05-29
Change Dlls to debug for Windows and MacOS

SahilTara:
Summary of changes
- change windows dlls to debug
- change mac dlls to debug
- Add an environment check for debug mode or production, and choose the sync path accordingly. Did this by adding an environment variable to the launch.json, and using that env variable within the launchExperience file.

The last commit contains any real code that is being committed. Everything else is auto generated!

Linux dlls will come separately since they seem to be problematic to remove.
Related issue:
#632


src/extension/.vscode/launch.json
```diff
@@ -10,6 +10,7 @@
       "type": "extensionHost",
       "request": "launch",
       "runtimeExecutable": "${execPath}",
+      "env": {"DEBUG": "yes"},
```
jcoc611-microsoft: Do you think it makes more sense for this one to be a boolean variable? (i.e. DEBUG=1 for debug mode, or DEBUG=0 or unset for production mode)?

SahilTara: Honestly it would make sense if numbers were allowed since 1 and 0 work like a boolean. In fact originally i tried doing true or false. But it said it had to be a string or null so I chose "yes" here. Do you think "1" would make more sense?

SahilTara: Im up for suggestions on different string possibilities as well such as "on", "enabled" etc

SahilTara: Now that I think about it "1" == true in javascript/typescript, also non null values also equal true in an if statement, do you think I could just set the "DEBUG": "1" here and then set the if to
```if (process.env.DEBUG) { .... } else {...}```?

jcoc611-microsoft: Oh now that you mention the type limitation there I think I'm okay either way. A different approach could be `NODE_ENV: "dev"` instead.

SahilTara: Done!

jcoc611-microsoft (approved):

jsondoo (approved):

piotrmark (approved):

jsondoo:
image

PR #640, opened 2019-05-29
Changed local linux DLLs to debug

### LINKED ISSUE #632: We need to change our local development DLLs to debug mode.

SahilTara:
Summary of changes:
- Changed local linux dlls to debug.

All files in this PR are auto generated.
Related issue:
closes #632


piotrmark (approved):

sibille (changes requested):
Can we close this one? Those dll s are out of date after bringing in PR https://github.com/microsoft/WebTemplateStudio/pull/678

Issue #623, opened 2019-05-27
Disable "Add Resource" button if not signed into Azure Services


xtinah-w:
Clicking on "Add Resource" crashes the wizard - needs to be disabled until the dev is logged in.

crutkas:
How did the wizard get into this state?  There should be a sign in button

crutkas:
Also what unit tests were added to catch this scenario

SahilTara:
This was an issue for a while it was a low pri item when we did it though. Probs some selenium tests would be one way.

PR #642, opened 2019-05-29
Disable add resource button if not signed in

jsondoo:
Issue #623
If the user is not signed in, the 'Add Resource' button should not be clickable. (right now it crashes the extension)
Also css changes so that if the button is disabled it shows the default cursor, if not it shows the pointer cursor.

jj


annielovesjs (approved):
Works perfectly when tested - mimicks the exact result showed in gif

jasmingoh (approved):
Works just as your described!

Tanya0609 (approved):

SahilTara (approved):

Issue #625, opened 2019-05-27
On the Summary page, change wording from "Welcome" to "Project Name"


xtinah-w:
Improves match between system and real world speak since "Welcome" isn't indicative of anything

PR #644, opened 2019-05-29
Changed 'Welcome' to 'New Project'

jasmingoh:
Resolving issue [#625](https://github.com/microsoft/WebTemplateStudio/issues/625)

Talked to Christina (designer on the team) about changing the "Welcome" text to "New Project". 

We looked at Windows Template Studio as reference, and found that they don't have a "Welcome" page, i.e. they start with the "Project Type" page. Before the Windows Template Studio extension pop up, the user is on a "New Project" window. Hence, we decided to make this change.

Here's the before and after:

Before:
BEFORE

After:
AFTER


jcoc611-microsoft:
Do we also need to change `ROUTES.WELCOME` to `ROUTES.NEW_PROJECT`? What about `src\client\src\containers\Welcome` to `src\client\src\containers\NewProject`? Might be a bit confusing to someone new coming in to the codebase

jasmingoh:
@jcoc611-microsoft Thanks for pointing that out! It make sense to change those filenames. I'll look around and see if there is any other relevant files.

crutkas:
Did you mean to close this branch?

jasmingoh:
I wanted to rename the branch because it did not make sense, but I didn't know it would close this one..

Issue #197, opened 2019-03-27
Add in Flask support


crutkas:

SahilTara:
@njindal239  is this done (I think it is)

Tanya0609:
Flask is in WebTS now, thus close this issue

PR #645, opened 2019-05-29
Create Flask mocks with a React Application

streetcoder123:
**Task**: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/27116

**Aim**: Bring Flask as an option for back-end framework

**Purpose of this PR:**
- This PR just creates some mock code for flask.
- It integrates react application as the front-end with Flask as the back-end similar to a project generated by web template studio
- This PR is using the sample data instead of azure services like CosmosDB. 

**Note**: Since this PR includes a boilerplate react code as well to work with Flask, I shall point the main areas to look for in this PR:
- The server folder that contains the flask application - server.py, constants.py and sampleData.py

**Steps to Run the Project**
- Make sure you have python (version 3.7) installed into the machine (https://www.python.org/downloads/)
- git checkout garage/t-nijind/flask-react-mocks
- cd mocks/flask-react-mocks
- yarn install (to install the dependencies)
- yarn start (concurrently starts the front-end react and back-end flask applications) 

I would really appreciate any feedback on the 'flask style of coding' because I am not very familiar with the same. When this gets merged, I shall go ahead and try to integrate CosmosDB into the mix (which would be added in another PR).

Finally, the main aim of getting these mocks approved is to make it easier to convert these into templates for Flask to be incorporated into WebTS,


SahilTara:
Quick question before I begin my review: since flask is just serving the backend here we could have any front end without changing our existing code right? i.e. if this was templatized we wouldn't have to redo react/angular/vue. I was a bit confused by the wording "Since this PR includes a boilerplate react code as well to work with Flask".

mocks/flask-react-mocks/README.md
```diff
@@ -0,0 +1,85 @@
...
+## File Structure
+
+The front-end is based on [create-react-app](https://github.com/facebook/create-react-app), and the
+back-end is based on [Express Generator](https://expressjs.com/en/starter/generator.html). The front
+end is served on http://localhost:3000/ and the back-end on http://localhost:3001/. During local
+development front-end and back-end communicate via a proxy defined in the package.json.
+
+```
+.
+├── server/                         - Express server that provides API routes and serves front-end
```
SahilTara: nit: mocks should contain the expected layout of flask here inside the server/ folder. This will make the templates easier for you.

mocks/flask-react-mocks/README.md
```diff
@@ -0,0 +1,85 @@
...
+3. Deploy the server folder to Azure App Service using the [Azure App Service Extension](https://marketplace.visualstudio.com/items?itemName=ms-azuretools.vscode-azureappservice).
+4. If a database is used, add the environment variables defined in .env to your Application Settings.
+5. Consider adding authentication and securing back-end API's by following [Azure App Service Security](https://docs.microsoft.com/en-us/azure/app-service/overview-security).
+
+Full documentation for deployment to Azure App Service can be found here: https://github.com/Microsoft/WebTemplateStudio/blob/dev/docs/deployment.md
+
+## File Structure
+
+The front-end is based on [create-react-app](https://github.com/facebook/create-react-app), and the
+back-end is based on [Express Generator](https://expressjs.com/en/starter/generator.html). The front
```
SahilTara: this should talk about flask.

mocks/flask-react-mocks/README.md
```diff
@@ -0,0 +1,85 @@
...
+├── .env                            - API Keys
+└── README.md
+```
+
+## Additional Documentation
+
+- React - https://reactjs.org/
+- React Router - https://reacttraining.com/react-router/
+- Bootstrap CSS - https://getbootstrap.com/
+- Express - https://expressjs.com/
```
SahilTara: link to flask

mocks/flask-react-mocks/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "MicrosoftWebTemplateStudio",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
```
SahilTara, line 5: probably won't need half of these dependencies look inside of the nodedefault template.json to figure out which dependencies can be removed.

streetcoder123:
@SahilTara That is correct. @annielovesjs is in the process of verifying it by integrating the code with Angular. However, the code is pretty much independent of what front-end framework you choose to run. You just need to plop the `server` folder with any front-end framework

mocks/flask-react-mocks/server/sampleData.py
```diff
@@ -0,0 +1,105 @@
+import datetime
+
+shortLoremIpsum = "Lorem id sint aliqua tempor tempor sit. Ad dolor dolor ut nulla mollit dolore non eiusmod Lorem tempor nisi cillum."
+longLoremIpsum = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Rhoncus dolor purus non enim praesent elementum facilisis leo vel. Risus at ultrices mi tempus imperdiet."
+
+sampleGridData = [
```
SahilTara, line 6: This is the sample data for master detail actually,
Grid is actually:
```
sampleData.gridTextAssets = [
  {
    description: shortLoremIpsum,
    header: "Company A",
    id: 1
  },
  {
    description: shortLoremIpsum,
    header: "Company B",
    id: 2
  },
  {
    description: shortLoremIpsum,
    header: "Company C",
    id: 3
  },
  {
    description: shortLoremIpsum,
    header: "Company D",
    id: 4
  },
  {
    description: shortLoremIpsum,
    header: "Company E",
    id: 5
  },
  {
    description: shortLoremIpsum,
    header: "Company F",
    id: 6
  },
  {
    description: shortLoremIpsum,
    header: "Company G",
    id: 7
  },
  {
    description: shortLoremIpsum,
    header: "Company H",
    id: 8
  },
  {
    description: shortLoremIpsum,
    header: "Company I",
    id: 9
  }
];
```

mocks/flask-react-mocks/server/sampleData.py
```diff
@@ -0,0 +1,105 @@
...
+    'status': "Shipped",
+    'shipTo': "Anna Bedecs",
+    'orderTotal': 800.0,
+    'orderDate': datetime.datetime.now(),
+    'id': 9
+  }
+]
+
+sampleData = {
+    'listTextAssets': [
```
SahilTara, line 100: this diverges a bit from the node data:
```
sampleData.listID = 3;

sampleData.listTextAssets = [
  {
    text: shortLoremIpsum,
    _id: 1
  },
  {
    text: shortLoremIpsum,
    _id: 2
  }
];
```
Was this intentional?

SahilTara (changes requested):
Will review server.py in a bit have to do something.

mocks/flask-react-mocks/server/server.py
```diff
@@ -0,0 +1,54 @@
...
+    listItem = {'_id': sampleData['listId'], 'text': data['text']}
+    sampleData['listTextAssets'].append(listItem)
+    sampleData['listId'] += 1
+    return jsonify(
+        listItem
+    )
+
+@app.route(CONSTANTS['ENDPOINT']['LIST'] + '/<int:id>', methods=['DELETE'])
+def deleteListItem(id):
+    listItemToRemove = [listItem for listItem in sampleData['listTextAssets'] if listItem['_id'] == id]
```
SahilTara: We know that the id is unique so why create a list? It also seems that this isn't used anywhere but for checking if a list item can be removed.

Maybe we should try something like this?
```py
for i in range(len(sampleData['listTextAssets'])): 
    if sampleData['listTextAssets'][i]['_id'] == id: 
        del sampleData['listTextAssets'][i] 
        return jsonify(....)
return jsonify(error here)
```

mocks/flask-react-mocks/server/server.py
```diff
@@ -0,0 +1,54 @@
...
+def getGrid():
+    return jsonify(
+        sampleGridData
+    )
+
+# MasterDetail Endpoint
+@app.route(CONSTANTS['ENDPOINT']['MASTERDETAIL'])
+def getMasterDetail():
+    return jsonify(
+        sampleGridData
```
SahilTara, line 50: Please see my message on the SampleData file.

SahilTara (approved):

mocks/flask-react-mocks/server/server.py
SahilTara: Unless you have a reason for this

SahilTara (changes requested):

SahilTara:
woah misclicked accidently approved, meant to request changes.

mocks/flask-react-mocks/README.md
streetcoder123: Do you mean I should make changes to the README based on the folder structure I have or are you suggesting a different folder structure for flask?

SahilTara: Readme based on the folder structure you have.

mocks/flask-react-mocks/server/server.py
```diff
@@ -0,0 +1,54 @@
...
+    return jsonify(
+        sampleData['listTextAssets']
+    )
+
+@app.route(CONSTANTS['ENDPOINT']['LIST'], methods = ['POST'])
+def addListItem():
+    data = request.data.decode('utf-8')
+    data = ast.literal_eval(data)
+    listItem = {'_id': sampleData['listId'], 'text': data['text']}
+    sampleData['listTextAssets'].append(listItem)
```
SahilTara: Feel like you should add items like so: sampleData['listTextAssets'] = [listItem, *sampleData['listTextAssets']]
or even an sampleData['listTextAssets'].insert(0, listItem) would be good because of the following bug:

added items and saw this Before refreshing: 
image

After refreshing:
image

SahilTara (changes requested):

SahilTara:
Fully tested this out with vue, works (so i'm 99% sure it will work with angular). Forgot the fact that you need to pip install flask for this to work. 

Wondering if we will be autogenerating venvs for the user or we assume they will do it?

mocks/flask-react-mocks/server/sampleData.py
streetcoder123: I looked at the grid component. It is actually using the property `gridTextAssets.shortDescription` instead of `gridTextAssets.description`. Hence, the above data would not work for a grid component. Although you are right that it never uses the `longDescription` attribute which is only being used in the masterDetail page.

mocks/flask-react-mocks/server/sampleData.py
streetcoder123: No, this was not intentional. I just put some random text to keep it simple but forgot to update it later. Shall update now.

mocks/flask-react-mocks/server/server.py
streetcoder123: Agreed!

streetcoder123:
Tested this out with Angular and it works as expected!

mocks/flask-react-mocks/server/sampleData.py
SahilTara: you know what this PR made me realized we are no longer using gridTextAssets in the nodes backend. I will be making a PR to remove this from all merging. Thanks for noticing this!

SahilTara: Or do you believe we should keep this sample data and instead change the frontend components?

mocks/flask-react-mocks/server/sampleData.py
SahilTara: Cool!

mocks/flask-react-mocks/server/sampleData.py
streetcoder123: Ideally, I think we should be changing the front-end components because that would make more sense for any person.

SahilTara: So you're saying we keep the sampleGridData that i provided (for node) and we change the frontend to use it rather than sharing the data between grid and master detail?

streetcoder123: Yeah, that is what I meant. However, I think for now, lets share the data between grid and master detail page. Changing this would mean changing a lot of things like the route in node.js and all front-end frameworks. I am not sure whether it would be an ideal thing to do now or should be incorporated once we have these frameworks (Vue and Flask) in. It is just an opinion.

What do you think about it?

UPDATE:
I discussed with the team. Let's first get the Vue framework in and then we can create a single PR to modify all the 3 front-end frameworks (React, Angular and Vue) to use sampleGridData instead of sharing with master detail.

SahilTara: Okay so I asked kai and he said clint wanted it to be shared between grid and master detail so im going to remove it from node in a seperate pr

mocks/flask-react-mocks/package.json
SahilTara: 'cookie-parser': '^1.4.3', 'debug': '~4.1.1', 'morgan': '^1.9.1', 'express': '^4.16.4', 'body-parser': '^1.18.3' @njindal239  these dependencies can be removed.

SahilTara: Actually remove everything that isn't the following: "{'bootstrap': '^4.3.1', 'classnames': '^2.2.6', 'react': '^16.8.4', 'react-dom': '^16.8.4', 'react-router-dom': '^4.3.1', 'react-scripts': '2.1.8'}",

streetcoder123: I am not sure if we would need `nodemon`. The front-end is booting up using react-scripts. So, I think it is not using nodemon anywhere.

SahilTara:
@njindal239 After the dependencies are removed I will approve this, I would try running this against pep and see if there's any problems that get flagged too if there are any. Style changes can be fixed in a separate PR though for the mocks.

mocks/flask-react-mocks/package.json
SahilTara: Yup you can remove Nodemon as well from the dev deps good call.

SahilTara (approved):

jcoc611-microsoft:
Quick question, maybe for @SahilTara . Are the `mocks/` supposed to be the output of WebTS (say in this case if you chose Flask + React)? If so why are they needed in the codebase (as opposed to only having the templates which will generate them)?

SahilTara:
Mocks/ are required in the codebase for quick and easy fixes in the future. Also for making the template creation process easier. Mocks are also easier to quickly iterate and get feedback on than templates. @jcoc611-microsoft

crutkas:
A few things here:

- what is the github issue for this.  
- The title of this is confusing, flask mocks with a react application?
   - Why did we pick flask versus something else
   - Is how this implemented is how a dev using flask would expect?  (can we link to a doc that can help me understand?)
- why isn't this in a dedicated branch versus a PR against dev

SahilTara:
I can answer some of these but good callout on lack of linked issue @crutkas.

- what is the github issue for this.
 It would be the preliminary for #197. 

- The title of this is confusing, flask mocks with a react application?
Agreed the title is confusing, anything you would suggest? This is meant to convey that these are flask mocks with a react front end. 
    - Why did we pick flask versus something else
    ASP.NET Core cannot be implemented as is,  since normally you would not have a server folder just for it, at least from all sources I've read. The react app would be inside the ASP.NET Core folder structure. For this we need a custom global post action which moves the client to inside of the ASP.NET structure. 
What could have been done is a simple api that is consumed but that is really forced and not how many articles . 
https://www.talkingdotnet.com/wp-content/uploads/2018/01/React-App-Structure-Comparison.png
Here is an example of such an article. Notice the "ClientApp Folder".

Now that leaves Django, Flask and Pyramid.

Django and flask are both super popular frameworks.
Here is a comparison between the two on google trends:
image

Edit: Noticed this was only for US here's worldwide.
image

Have a look at this for a bit more info though.
https://coderseye.com/django-vs-flask

I feel like one of them is easier to implement than the other I think the team can elaborate more for this point.

- why isn't this in a dedicated branch versus a PR against dev
Would you rather it all be in a feature branch until all of flask is ready? its a matter of preference really. There's no harm in putting it inside of dev. In fact it actually increases exposure in my opinion. For feedback that is.  This doesn't affect template generation so I feel like its fine, since we won't have any half built templates inside of the extension.
Team can elaborate more.

jsondoo:
>  Are the `mocks/` supposed to be the output of WebTS (say in this case if you chose Flask + React)?

@jcoc611-microsoft  No, mocks/ are complete separate. It's more for us (the team) to work on and develop, and once we feel that it's ready we will move it into templates/

jsondoo:
> * what is the github issue for this.

#197 , we should probably update the issue

> The title of this is confusing, flask mocks with a react application?

agreed

>   * Why did we pick flask versus something else

We talked about this on Teams and got approval at the Sponsor Sync from Mike while you were OOF. One of our archetype was student developer or hackathon participants, and the hackathon submission data showed that Flask was way more popular over other Python frameworks and Ruby on Rails, so we felt it was good fit for our target customers, and something they would really want. Annie can add more to this since I can't find the data scraped right now. Flask is also a lot lighter so it would be great for prototyping which is what our product is for (but my flask knowledge is limited so maybe someone can add to this) We can definitely look into the other frameworks, we are just prioritizing Flask for now.

> * why isn't this in a dedicated branch versus a PR against dev

Adding to what Sahil said, it would be a lot easier for us to develop the mocks if it was on dev versus on another feature branch. It was also what was done with React and Angular mocks previously and it's completely separate from templates so I don't think it's an issue. Maybe you can elaborate on why you would prefer a branch instead? @crutkas

streetcoder123:
@SahilTara and @jsondoo  Thank you for going in such detail for answering questions.

@crutkas I will also try to explain your questions:

1. As Sahil pointed out, this PR would be a preliminary to GitHub issue #197. However, this also brings me to a question. For every feature/task that we decide to work on (task created in Azure Devops), is a corresponding issue needed to be created on GitHub?

2. 
- I agree that the title of the PR is a bit misleading. It seems as if we are creating flask mocks which work only with React Application. However, the idea here is to create flask mocks. React is randomly chosen as the front-end to see application behavior. However, these mocks have been verified against all of React, Angular and Vue for its functionality.
- For Flask as being chosen as the back-end framework, @SahilTara has a pretty good description. Also, points made by Jason pretty much sums up our research on coming up to this framework choice.
- **Is how this implemented is how a dev using flask would expect?**
To understand how a developer would write a Flask Application, I looked up some examples on GitHub that had similar structure to this one for the server. However, those applications involved creating templates with Flask (a feature that flask offers) but that is not the motive of our application since we have dedicated front-end frameworks to handle those. Also, understanding the best practices while writing Flask applications is our top priority for next Sprint. We have a meeting arranged with an "expert" tomorrow as well and will try to get more insights next week.
This will serve as a validation of the code style that we use while writing the Flask server.

3. I strongly believe that we should have the mocks in dev for the reason as described by Sahil that it makes much more easier to verify changes in future. For instance, if we want to make some changes to flask codebase in future, it makes it really easy for developers to test code on mocks before making final change on the templates. Verifying by making changes directly on templates is time-consuming since we would have to generate the project to test out changes.
Also, as mentioned by Jason, we already have a completely segregated folder called `mocks` in dev where we have react and node mocks as well. 

Let me know if any of the above seems unclear.

jcoc611-microsoft:
> Django and flask are both super popular frameworks.
> Here is a comparison between the two on google trends

I'm all in with the team's decision to go with Flask. However, I think google trends has too much noise to be useful for deciding, as "flask" and "django" are not exclusively the names of web frameworks but also of other popular things. Here is a better source of info, the StackOverflow yearly survey (out of 63,585 responses, for year 2019):
image

As they are very close to each other, I think going either way is fine.

SahilTara:
Yup of course was just a quick example i could pull out.

sibille:
For wints we do mocks in separate repos, linked in the issue. This makes it easier to share them with other people (you don't have to clone the whole repo to look at the flask mocks.) Once they are converted into templates we delete them. For evolutions we create a new mock from the templates, takes you a second and ensures you're in sync with the templates (as single source of truth.)

jcoc611-microsoft:
> For wints we do mocks in separate repos, linked in the issue. This makes it easier to share them with other people (you don't have to clone the whole repo to look at the flask mocks.) Once they are converted into templates we delete them. For evolutions we create a new mock from the templates, takes you a second and ensures you're in sync with the templates (as single source of truth.)

This makes a lot more sense to me. Coming into the WebTS repo for the first time, I was really confused about the difference between mocks and templates. I assumed some automated build step converted from mocks to templates or vice versa.

jsondoo:
> For wints we do mocks in separate repos, linked in the issue. This makes it easier to share them with other people (you don't have to clone the whole repo to look at the flask mocks.) Once they are converted into templates we delete them. For evolutions we create a new mock from the templates, takes you a second and ensures you're in sync with the templates (as single source of truth.)

Yea, this actually makes sense to me. I would be open to having a separate repo or branch just for mocks and removing the `mocks/` folder from this repo

streetcoder123:
@sibille That is a good point. Mocks would get out of data soon as we make changes to the templates in future. Maintaining mocks as well as templates adds more overhead. I would agree that we should add the mocks to a different repo. This should also be followed by removing the mocks folder from the dev since react mocks are already out of date.

SahilTara:
@njindal239 should we close this PR?

streetcoder123:
@SahilTara Yeah, I will close this PR after sponser sync today. Looking to get a final confirmation from Clint on where are we going with the mocks.

streetcoder123:
I am closing this PR since we decided for mocks to be in dedicated branch rather than being in the dev code base.

PR #647, opened 2019-05-30
Base composition templates for vue

SahilTara:
Summary of changes:
Base composition templates for vue. These need to be PR'd before the pages to allow testability of all PR branches separately. This PR isn't testable until the pages PR.

Should be reviewed after #617 

Just hit f5 and launch webts then hit generate! Dlls are in for windows and mac!
Related issues:
#199

Issue #199, opened 2019-03-27
Add in Vue support


crutkas:

SahilTara:
I can add this feature if need be, just assign it to me after mentioning me.

crutkas:
if you want it, take it so it gets to feature parity of the React support

SahilTara:
Okay cool I'll work on it this Friday, will be based off of https://github.com/microsoft/WebTemplateStudio/tree/t-kabail/vue-mocks

Would you want me to make it preview only like we've done with Angular?

crutkas:
sure

SahilTara:
@crutkas I have lined up PRs ( will come after merging of certain PRs but the branches are ready) for this issue can you add the in progress tag since I don't have permissions to do this anymore. :smile: Hope you're ready for vue! I have the preview tag on it for now, but I'm fairly certain that it will be at feature parity with react once all prs are in.  So once all PRs are in we will need to do a similar process to what we are doing with angular, and that is to verify we have feature parity and then turn the flag off.

crutkas:
Will take a look today / tonight (on vacation 😁)

SahilTara:
Oh oops did not read this message at all until now! Sorry, enjoy your vacation!

crutkas:
who from the vue community has reviewed these?

Tanya0609:
Since we have created another issue to log all the feedback from Vue experts and tracking the progress, thus closing this issue #978

PR #631, opened 2019-05-28
Base templates for vue

SahilTara:
Summary of changes:
Implements the base of Vue templates.  Vue is set under a preview flag. This is what is always generated no matter what is selected.

To test PR will need to package locally.

Related issue:
#199

PR #648, opened 2019-05-30
Blank Page for vue

SahilTara:
Summary of changes:
Implements the blank page for vue. You should now be able to generate a project with a blank page!

Should be reviewed after #647 

Just hit f5 and launch webts then hit generate! Dlls are in for windows and mac!
Related issues:
#199


Tanya0609 (commented):
Try to use the change to generate a blank vue page, it works. Code looks good as well.

Will approve this PR once #631 is merged and this gets rebased

templates/Web/_composition/Vue/Page.Vue.AddConstantsFile/.template.config/template.json
```diff
@@ -0,0 +1,30 @@
+﻿{
+    "author": "Microsoft Community",
+    "classifications": [
+      "Universal"
+    ],
+    "name": "Page.Vue.AddConstantsFile",
+  "tags": {
```
jcoc611-microsoft: Should indentation be fixed here?

templates/Web/_composition/Vue/Page.Vue.AddHome/.template.config/template.json
```diff
@@ -0,0 +1,33 @@
+﻿{
+    "author": "Microsoft Community",
+    "classifications": [
+      "Universal"
+    ],
+    "name": "Page.Vue.AddHome",
+  "tags": {
```
jcoc611-microsoft: Same question about indentation

templates/Web/_composition/Vue/Page.Vue.AddConstantsFile/.template.config/template.json
Tanya0609: I think the formatting issue is propagate from the same file in React. Also the other PRs are all interdependent on each other. I am concerned if change here will create merge conflicts for the remains of the PRs. Maybe we can have a separate PR to fix those formatting use Prettier together? Or you guys think it will be good to fix this one first?

SahilTara: Yup this can be fixed. Actually all pages PRs are independent (Once vue base compositions is in my other PRs are independent).

templates/Web/_composition/Vue/Page.Vue.AddHome/.template.config/template.json
SahilTara: Yup you're right I should fix this.

SahilTara: Actually I noticed this is an issue for almost all of my template files in the repository. Do you think we should make a PR fixing formatting of all of these. I usually turn format on save off for templating because it can actually break certain templates (removes macros or edits certain lines). But we can make a giant PR fixing all template.jsons if you wish (which I think would be the better option (since this does not affect users)). Your call here though because I can fix it for all pages PRs right now.

templates/Web/_composition/Vue/Page.Vue.AddConstantsFile/.template.config/template.json
jcoc611-microsoft: In that case yeah we can address the indent issue independently, thanks!

jasmingoh (approved):
Was able to generate blank pages and the code looks good!

PR #652, opened 2019-05-30
Vue Functions Composition

SahilTara:
Summary of changes:
Implements the functions merging for vue. Selecting Azure functions with vue selected will now add an excerpt in the read me about azure functions. Mimics behavior of React+Angular.

Should be reviewed after #647

Just hit f5 and launch webts then hit generate! Dlls are in for windows and mac!
Related issues:
#199


jcoc611-microsoft:
Despite the name, this has a bunch of compositions for Vue other than Azure Functions right? Still trying to understand the whole structure of the code

Edit: Closed by accident

SahilTara:
Ah this PR shouldn't be reviewed yet. No its just for azure functions actually. Honestly I feel like this entire template should be in its own location now that you mention it.

SahilTara:
I've done the same for angular + react as well so I should probably abstract it out into its own composition. Thanks!

jcoc611-microsoft:
> Ah this PR shouldn't be reviewed yet. No its just for azure functions actually. Honestly I feel like this entire template should be in its own location now that you mention it.

Now that I think about it, do you think you could merge all the Vue-related stuff into a single PR? I think it would greatly help us on our end with reviewing the changes

SahilTara:
Oh really? I split it all up because I thought that would make reviewing and such a lot easier. Will do so.

SahilTara:
PR might be a bit big though.

SahilTara:
Closed in favour of #676

PR #653, opened 2019-05-30
Changed 'Welcome' to 'New Project'

jasmingoh:
(Copied over from PR #644 - changed branch name to something that makes more sense)

Resolving issue [#625](https://github.com/microsoft/WebTemplateStudio/issues/625)

Talked to Christina (designer on the team) about changing the "Welcome" text to "New Project". 

We looked at Windows Template Studio as reference, and found that they don't have a "Welcome" page, i.e. they start with the "Project Type" page. Before the Windows Template Studio extension pop up, the user is on a "New Project" window. Hence, we decided to make this change.

Here's the before and after:

Before:
BEFORE

After:
AFTER


jasmingoh:
Thanks to @jcoc611-microsoft 's [comment](https://github.com/microsoft/WebTemplateStudio/pull/644#issuecomment-497105271) in the other PR (now closed, because I renamed the branch name)

jcoc611-microsoft (approved):

crutkas (approved):
Why do we have default messages. This should default to en-us or a default file versus having it hard coded. This should be logged as a new issue

piotrmark:
> Why do we have default messages. This should default to en-us or a default file versus having it hard coded. This should be logged as a new issue

I may be wrong, but I think en-us is auto-generated from default messages.

PR #656, opened 2019-05-30
Warning in Browser Console In React List Template

### LINKED ISSUE #638: Warning in Browser Console In React List Template
Steps to reproduce the warning:
- Generate a React application in Web Template Studio
- Do 'yarn install' to install the dependencies and 'yarn start' to run the application
- Go to 'localhost:3000/list' to see the list page and open browser console
- Type something into the form to submit a new list item (you will see the following warning in the console) 

Capture

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/27265
Github Issue: #638 

**Cause of Issue**
This warning is caused when we try to set state for an attribute whose state has not been initialized. The change event for the form should have actually been handled by the form as it has nothing to do with the parent list component.

**Changes**
Make form a class component so that we can incorporate the change event logic within that component. 

**Result**
No warning appears on the console when we type something into the list form.

closes #638


templates/Web/Pages/React.List/src/components/ReactList/ListForm.jsx
```diff
@@ -1,20 +1,41 @@
-﻿import React from "react";
```
jcoc611-microsoft, line 1: Does it make sense to keep the previous style of `;` at every line end?

templates/Web/Pages/React.List/src/components/ReactList/index.jsx
```diff
@@ -15,7 +15,6 @@ export default class ReactList extends Component {
 
     this.handleWarningClose = this.handleWarningClose.bind(this);
     this.handleDeleteListItem = this.handleDeleteListItem.bind(this);
-    this.handleChangeInputText = this.handleChangeInputText.bind(this);
```
jcoc611-microsoft, line 18: I think you also should remove the prop from line 119

piotrmark (changes requested):
Like Juan said, you changed the props specification of ListForm, so now it reads only onAddListItem.
So, you need to remove onChangeInputText and textField from each declaration of <ListForm ...> tag.
And propagate these changes up (i.e. text field would no longer be necessary in ReactList state).

I'd recommend using PropTypes to avoid manual checks like that in the future:
https://blog.logrocket.com/validating-react-component-props-with-prop-types-ef14b29963fc

templates/Web/Pages/React.List/src/components/ReactList/ListForm.jsx
streetcoder123: That makes sense. I just out of habit forget to add those. Thanks for pointing that out.

streetcoder123:
@piotrmark Thanks for the feedback. I have made the changes. For adding the proptypes, I think that should be done across all components. We as a team can think about that and come up with a different issue solely purposed to add prop types. Nevertheless, a great idea.

SahilTara:
Hey @njindal239 noticed you committed the mstx file. You don't need to package anymore!

streetcoder123:
@SahilTara Yeah, I realized that just after committing.

templates/Web/Pages/React.List/src/components/ReactList/ListForm.jsx
```diff
@@ -1,20 +1,41 @@
...
-        placeholder="Add text here..."
-        aria-label="Add text here..."
-      />
-      <button type="submit" className="btn btn-primary ml-2">
-        Submit
-      </button>
-    </form>
-  );
-}
+export default class ListForm extends Component {
```
SahilTara, line 3: quick question why did you do the move from a functional component to a class component?

SahilTara: If it was because of the state initialization of the text field isn't this actually a good opportunity to start migrating slowly towards react hooks? 

Instead we could do something like this
```js
const [textField, setTextField] = useState("");
```

textField would be the readonly value of textField, while setTextField() would set the value of the said text field.

SahilTara: Any thoughts @piotrmark?

jcoc611-microsoft: React Hooks is kind of a [controversial](https://news.ycombinator.com/item?id=18302162) addition to React. We would have to think about it more from the perspective of shipping (opinionated) best-practice templates.

SahilTara: All good just wanted to have some input on if we should or not. Mainly because one of the experts we reviewed with said it would be a nice to have the entire thing redone with react hooks. But as you said it would be best if we spend more time thinking about its addition or lack of.

piotrmark: Yeah, React hooks is a super fresh feature (maybe it's not experimental any more, but probably not well established in community yet).
If we want to move to hooks, IMO it would need to be one big switch instead of changing one component at a time as probably we don't want inconsistent templates.
Regardless, it's out of scope for this task which is about eliminating the warning about input management.

templates/Web/Pages/React.List/src/components/ReactList/ListForm.jsx
piotrmark: It should be Prettier's job :)

piotrmark (approved):

templates/Web/Pages/React.List/src/components/ReactList/ListForm.jsx
SahilTara: It is definitely super fresh came out in 16.8 I think. We are using it on our client side right now!

Yeah I realized that after the fact that it would be a better choice if we did a full overhaul with hooks if we did choose to do it. You are completely correct about the inconsistencies created with not doing such. 

Yup definitely out of scope just the switch was a good time to bring up the question.

SahilTara (approved):

Tanya0609 (approved):
code looks good and the warning is not showing anymore

PR #663, opened 2019-05-31
Allowed start scripts to be decoupled from the backend

### LINKED ISSUE #655: Script tag to start the application depends on both frontend and backend
Currently, the frontend configures the script for `yarn start`. For example react configures it as:
`concurrently \"react-scripts start\" \"nodemon server/server.js\"`
Similarly, for vue we have:
`concurrently \"vue-cli-service serve\" \"nodemon server/server.js\"`

So, the Front-end assumes that the backend would always be Node.js. This would create issues when we start adding more backend frameworks. We could have a composition template that adds into "scripts" section after detecting the frontend + backend selection.

SahilTara:
Summary of changes:

- Moved start scripts from base project templates into global postactions.
- Postactions filtered by frontend + backend combo to insert the appropriate start script.

closes #655


templates/Web/_composition/LaunchScript/Launch.AngularNode.AddScript/package$Launch.AngularNode.AddScript_gpostaction.json
```diff
@@ -0,0 +1,5 @@
+  "scripts": {
+//{[{
+    "start": "concurrently \"ng serve --port=3000 --o --proxy-config proxy.config.json\" \"nodemon server/server.js\"",
```
jcoc611-microsoft: Not super familiar with the templating system, but would it be possible to do this instead:

```
"start": "concurrently {[{\"ng serve --port=3000 --o --proxy-config proxy.config.json\"}]}",
```

Basically, decoupling frontend from backend scripts and thus reducing the number of `Launch.<Front><Back>.AddScript` templates?

SahilTara: Honestly I wish this was possible as well. Right now at least to my knowledge, Inline merging is not possible.

SahilTara: True decoupling would be as you suggested with something like  {[{\"ng serve --port=3000 --o --proxy-config proxy.config.json\"}]} however with the current capabilities of the core I don't think this is possible.

Actually I think I have an idea for something that might be possible I'm going to try it out.

SahilTara: What if we had something like this?

```
    "start": "concurrently \"npm run start-frontend || yarn start-frontend\" \"npm run start-backend || yarn start-backend\"",
    "start-backend": "nodemon server/server.js",
    "start-frontend": "react-scripts start",
```

SahilTara: This way I could merge in the start-frontend and start-backend scripts.

jcoc611-microsoft: That works actually! Nice one

Tanya0609: @SahilTara are you gonna update your PR with the code you guys mentioned in the conversation or you are going to stick with your original change?

Reviewed the original change, I am able to start my generate templates

templates/Web/_composition/LaunchScript/Launch.ReactNode.AddScript/package$Launch.ReactNode.AddScript_gpostaction.json
```diff
@@ -0,0 +1,5 @@
+  "scripts": {
+//{[{
+    "start": "concurrently \"react-scripts start\" \"nodemon server/server.js\"",
+//}]}
+  },
```
Tanya0609: It seems like this line ("start": "concurrently....) will get inserted into the **scripts** object inside **package.json** of generated templates.

Just two learning questions, 1) how does coreWebTemplate know which file it should insert code into and 2) which file it should grab the line it is going to use for insertion?

templates/Web/_composition/LaunchScript/Launch.AngularNode.AddScript/package$Launch.AngularNode.AddScript_gpostaction.json
SahilTara: Yeah will be implementing these changes, have to do an assignment first though

templates/Web/_composition/LaunchScript/Launch.ReactNode.AddScript/package$Launch.ReactNode.AddScript_gpostaction.json
SahilTara: it knows due to the file name notice "package_postaction.json", or in this case `"package$<some id>_gpostaction.json"`. The difference between the two is that package_postaction will be merged after the page/project template is merged but gpostaction will be merged after all other _postactions and templates are merged. So they are like a final merging step, this is especially helpful for guarantee of a certain format/existence of a certain file that may be added in a _postaction.

SahilTara: To my knowledge it searches for the `_postaction/$<id>_gpostaction` tags, and then it determines the original file it needs to insert into by removing the tag, so for our case it would be "package.json". Does this make sense?

templates/Web/_composition/LaunchScript/Launch.AngularNode.AddScript/package$Launch.AngularNode.AddScript_gpostaction.json
SahilTara: Hey @jcoc611-microsoft I have implemented the changes we discussed.

jcoc611-microsoft (approved):

templates/Web/Projects/AddPackageJsonFullStackDefault/.template.config/template.json
```diff
@@ -0,0 +1,37 @@
...
+  "sourceName": "wts.AddPackageJsonFullStackDefault",
+  "preferNameDirectory": true,
+  "PrimaryOutputs": [],
+  "symbols": {
+  },
+  "postActions": [
+    {
+      "description": "Add Base Start Scripts",
+      "manualInstructions": [],
+      "actionId": "CB387AC0-16D0-4E07-B41A-F1EA616A7CA9",
```
Tanya0609, line 28: what is the action id? is it save to put it inside GitHub?

Tanya0609 (commented):
Code change works for generating all three frontend frameworks. Once the conversion been resolved, I will approve this PR.

templates/Web/Projects/AddPackageJsonFullStackDefault/.template.config/template.json
```diff
@@ -0,0 +1,37 @@
...
+      "args": {
+        "dict": "{'start': 'concurrently \"yarn start-frontend || npm run start-frontend\" \"yarn start-backend || npm run start-backend\"'}",
+        "key": "scripts",
+        "jsonPath": "package.json"
+      },
+      "continueOnError": false
+    }
+  ]
+}
```
streetcoder123, line 37: @SahilTara That's a good idea to make start script to another file that is independent of the front-end and back-end framework. I think it would also be nice to have the 'concurrently' dependency being added in this file because it would only be used when we run a full stack application.
Independently, the front-end and back-end frameworks should not required access to 'concurrently' package.

What do you think about it?

templates/Web/Projects/AddPackageJsonFullStackDefault/.template.config/template.json
SahilTara: Completely safe. This id is used by corets to do the right custom post action. It is specifying that i want to do a json addition. Thanks for the concern but no it is not a private key haha, has nothing to do with access or anything. Just the unique identifier for the post action we want. There's a whole list on corets for the custom postactions.

templates/Web/Projects/AddPackageJsonFullStackDefault/.template.config/template.json
SahilTara: Its possible if you want i can add it

streetcoder123: I think we should add it as it would make more sense for that package to be here :)

templates/Web/Projects/AddPackageJsonFullStackDefault/.template.config/template.json
Tanya0609: thank you for the explanation. I will approve the PR

Tanya0609 (approved):

templates/Web/Projects/AddPackageJsonFullStackDefault/.template.config/template.json
streetcoder123: We shall incorporate this in another PR.

streetcoder123 (approved):

PR #670, opened 2019-06-01
changed React navbar links from anchor tags to Link components

### LINKED ISSUE #641: React NavBar component using anchor tags instead of React Router Links
The NavBar component is using anchor tags as navigation aids to the different pages, which forces unnecessary page refreshes. Using React Router's built-in Link component would be a simple fix for this. I'd be happy to update this if it's the best path forward!

scott-ammon:
Replaced anchor tags for NavBar with react-router-dom's Link component to prevent page refreshes on route navigation.

closes #641


templates/Web/Projects/ReactDefault/src/components/NavBar/index.jsx
```diff
@@ -7,12 +8,12 @@ export default function NavBar() {
   return (
     <React.Fragment>
       <div className={styles.skipLink}>
-        <a href="#mainContent">Skip to Main Content</a>
+        <Link to="#mainContent">Skip to Main Content</Link>
```
SahilTara: I think that this line should remain as: 
```html
<a href="#mainContent">Skip to Main Content</a>
```

I feel like the Link here will break the skip link. Thoughts?

mocks/react-mocks/src/components/NavBar/index.jsx
```diff
@@ -7,25 +8,25 @@ export default function NavBar() {
   return (
     <React.Fragment>
       <div className={styles.skipLink}>
-        <a href="#mainContent">Skip to Main Content</a>
+        <Link to="#mainContent">Skip to Main Content</Link>
```
SahilTara: Feel like this should remain as:
```html
<a href="#mainContent">Skip to Main Content</a>
```
Think this breaks skip link used for accessibility. Thoughts?

templates/Web/Projects/ReactDefault/src/components/NavBar/index.jsx
scott-ammon: Links do end up rendering an anchor tag in the DOM, but good point - i'm not sure how it treats skip links so i'll revert it.

mocks/react-mocks/src/components/NavBar/index.jsx
scott-ammon: Links do end up rendering an anchor tag in the DOM, but good point - i'm not sure how it treats skip links so i'll revert it.

SahilTara: Yeah its more so hashed links in general. Reverting fixed it! Will do another quick glance but everything seems to look good!

SahilTara (dismissed):

mocks/react-mocks/src/components/NavBar/index.jsx
```diff
@@ -10,22 +11,22 @@ export default function NavBar() {
         <a href="#mainContent">Skip to Main Content</a>
```
Tanya0609: Will the skip to main content link need to be updated as well?

SahilTara: No, as I mentioned in the above comments to Scott, if it is updated it breaks the skip link.

Tanya0609: sorry did not see the conversation. If so, we won't need to change the skip the main content for Angular neither right? Cause I am making similar changes for Angular.

Tanya0609: Also, how did you test accessibility?

SahilTara: Press tab @Tanya0609  you should see a pop up with skip to main content, press enter if it doesn't take you to main content, something broke.

jsondoo (changes requested):
Hey @scott-ammon, i gave this a try and i get an error saying 
`You should not use <Link> outside a <Router>`

Maybe try a router around the links?

scott-ammon:
Hey @jsondoo, I do not see any React console errors when I run mine locally. There is the BrowserRouter wrapped around App, which should handle the routing. I normally add a Router wrapping the Switch component since Routers can only have one child component, so I don't believe we can wrap all the navbar links in a Router.

streetcoder123:
The current version of dev does not have Navbar wrapped inside BrowserRouter. The change was made to make unit testing on generated application easier. However, since we will now be using React Router Links instead of anchor tags in Navbar, BrowserRouter needs to enclose Navbar as well. I will make a PR to make that change. With this change, I will also have to wrap the test file for App component within Browser Router to work. Once that change gets merged, this PR, I think will be good to go.

scott-ammon:
Ah that makes sense, thanks for finding that. Sounds good!

streetcoder123:
@scott-ammon You have a merge conflict. This conflict is due to the reason that mocks folder has been permanently removed from the `dev` branch since its contents were out of date. You will need to undo your changes from the react-mocks to solve the conflict.

streetcoder123 (approved):
This works good!
@jsondoo You would probably need to resolve your comments as well once you verify this.

jsondoo (approved):

jsondoo:
@scott-ammon looks good!

Issue #665, opened 2019-06-01
Angular navigation not using routerLink


xWiiLLz:
https://github.com/microsoft/WebTemplateStudio/blob/f5cf75d037c12c910719196da01e0f5ef811435c/templates/Web/_composition/AngularJS/Page.Angular.AddNavItem/src/app/components/NavBar/navbar.component_postaction.html#L5

The navigation links are not using routerLink, which makes them reload the entire page rather than leveraging angular's router.

Should be changed to `routerLink="wts.ItemName"`.

SahilTara:
Seems like we did the same thing in angular as we did in react! Good call out! Thankfully we didn't make the same mistake in Vue.

PR #675, opened 2019-06-04
angular nav not using routerLink

Tanya0609:
This change is to address1 of the generated angular code feedback from GitHub Issue: #665
Task:https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/WTS%20Team/Intern%20GitHub/Y19-VS/03

In Angular <a> tag should use "routerLink" rather than "href"

**How to test:**
1. generate an angular project with at least any two pages are selected 
2. start your application (npm install, npm start)
3. click links under nav bar should redirect to corresponding page. **(expected)**


SahilTara:
Do you think you (@Tanya0609)  or someone else could investigate a way to get the skip link working, in a different PR? Since it seems that its not working since before this PR as well. Think it has to do with href not being {{skipLinkPath}} or something of the sort. Even with that fix though with routerLinks it fails but with hrefs it passes, so there's probably a way to get it working with routerLinks. Just shouldn't have to deal with it if its already broken before this pr when this pr isn't targeted towards that.

SahilTara (approved):
Everything works as it should except skip links but it seems they didn't work before this PR either. Will have to investigate this

streetcoder123 (approved):

Tanya0609:
@sahilTara I will log skip link as a different GitHub issue and fix it in next spring

jsondoo:
> next spring

that sounds really late Tanya hhaha

jsondoo (approved):

Tanya0609:
It is a typo LOL next sprint**

piotrmark (approved):

Issue #681, opened 2019-06-04
Verify generated Angular templates code follows the best practice


Tanya0609:
Our team is trying to release Angular out of preview mode. We would like to have Angular experts and experienced users to provide some feedback on our generated Angular templates code to see if it follows the best practice. 

Some feedback has already been provided, if there are any more critical feedbacks or potential improvements, please feel free to add it in this issue.

manekinekko:
Hi, here are few quick notes (from interal discussion with @Tanya0609 ):
- [x] the content of the src/constants.ts should be moves to the environment files under src/environments
- [x] the files under src/app/images should be moves to the assets folder src/assets and use the MatIconModule to register and use the SVG assets (see https://material.angular.io/components/icon/api)
- [x] you may wanna move the WarningComponent under src/app/shared/warning/ (to be created)
- [x] I'd rename src/app/components to src/app/app-shell
- [x] all folders and symbol names should be in lower kebab case, eg, Master_Details.module.ts --> master-details.moduke.ts
- [x] I'd probably move the interface definitions from the components to their related services, eg, IGridTextItem in src/app/components/Grid/GridBox/grid-box.component.ts should be moved to src/app/components/Grid/grid.service.ts
- [x] make sure you run the ng lint command to fix  all formatting issues (if any). You may wanna make also sure the app is building in prod mode.
- [x] We released Angular version 8 yesterday, you may wanna give it a shot 😉
- [x] I was also wondering if you are planning to add some tests (if that is relevant)?

Let me know if you need anything else.

manekinekko:
Internal question from @Tanya0609 
> I have a question towards “- I'd rename src/app/components to src/app/app-shell”.  I used Angular CLI to generate an Angular app and use ng generate to generate more components. It seems each component is a folder under /app, rather than being group into another folder /app/app-shell or /app/components.

When you generate symbols using the CLI, by default everything will be generated at the root of `src/app` and the CLI will use the `app.module.ts` as the host module. We don't force any folder architecture in the CLI. 

The "app-shell" naming is just a suggestion I gave—sorry I should have made this explicit. I just noticed that all of your components and modules are all somehow related to the UI shell. "app-shell" is a common naming convention. But your are free to name it differently. 

If you want to specify a different folder, just use the desired folder name: `ng g c subfolder/component-name`. You can also provide other flags when generating symbols (see [angular.io/cli/generate](https://angular.io/cli/generate)).

[...]

PR #684, opened 2019-06-05
Fix Lint Errors in Angular

### LINKED ISSUE #683: Fix Lint Errors in a generated Angular Project
If we run 'yarn lint' on a generated angular application, we run into various lint errors. We need to fix those.

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/27724
Github Issue: #681

**Changes**
- It fixes various lint issues in different parts of angular application

**Note**: There is still one lint error remaining that would be handled by in another PR by @Tanya0609 as it is related to naming conventions of files.

**Procedure to check**
- Generate an Angular Application
- Inside the application install dependencies using: `yarn install`
- Run `yarn lint` to check there should be only 1 lint error


Tanya0609 (approved):
generated an angular project with all four type pages selected. Generated app works as usual. 
I did yarn install and yarn lint. All of the lint errors are gone except one that will be addressed in a different PR.

templates/Web/Pages/Angular.List/src/app/app-shell/AngularList/ListForm/list-form.component.ts
```diff
@@ -5,20 +5,15 @@
...
+  textField: string;
   @Output() inputTextEventEmitter = new EventEmitter<string>();
-  get textField() {
-      return this._textField;
-  }
-  set textField(textInput) {
-      this._textField = textInput;
-  }
+
   constructor() { }
```
jsondoo, line 11: How come this closing bracket isn't on a new line like the other ones? Or is constructor a special case

jsondoo (approved):
Verified lint is passing

templates/Web/Pages/Angular.List/src/app/app-shell/AngularList/ListForm/list-form.component.ts
streetcoder123: I don't think there is something special with constructor. But, this is normally the pattern I see for methods with no content.

templates/Web/Pages/Angular.List/src/app/app-shell/AngularList/ListItem/list-item.ts
```diff
@@ -5,6 +5,7 @@
   templateUrl: './list-item.component.html'
 })
 export class ListItemComponent implements OnInit {
+  // tslint:disable-next-line
```
jcoc611-microsoft, line 8: Quick question, why is this needed?

streetcoder123: Good question! So, the backend returns a list item that has attributes `text` and `_id`. When we make HTTP request to get the list data from the backend, it is assumed to be of type 'IListItem'. We cannot/should not change the data returned from the backend and if we modify the 'IListItem' in front-end to have `text` & `id` instead of `text` & `_id`, the data returned from the backend would break.

So, in that case, I thought we should keep the attribute `_id` as is and disable the lint for this one. Do you have any other suggestions?

jcoc611-microsoft: It sounds reasonable!

jcoc611-microsoft (approved):

Issue #590, opened 2019-05-16
Add more unit tests for client


sibille:
Would be nice if we can increase the code coverage with some useful tests. 

And more taking into account that most of them are commented out. See issue #1709

jcoc611-microsoft:
@jsondoo is this one done?

streetcoder123:
For this one, the testing environment and the build pipeline has been set. However, only 1 actual test has been written.

jsondoo:
@sibille can this be closed?

jsondoo:
Currently there are two unit tests, one for App.tsx and one for Card.tsx. We added a new command to package.json `test: react-scripts test --coverage` which uses the react-scripts Jest testing framework. This command is also run in our build pipeline

PR #688, opened 2019-06-05
Set up testing for client code

jsondoo:
Set up for #590 

- configure Jest
- Fix currently existing tests
- Add test for Card as a reference (we don't want to test connected components so made it a named import)
- Adds enzyme as a dependency (for shallow rendering) https://facebook.github.io/create-react-app/docs/running-tests we can add more libraries later if needed

**For testing**
Go to `src/client` and run `CI=true yarn test` to run all the tests and see the coverage

This will also be integrated to ADO build pipeline as well


src/client/package.json
```diff
@@ -50,7 +50,7 @@
     "build": "yarn manage:translations && node ./scripts/build-non-split.js",
     "postbuild": "rm -rf ../extension/react && mv build ../extension/react",
     "manage:translations": "node scripts/manageTranslations.js",
-    "test": "react-scripts test",
+    "test": "react-scripts test --coverage",
```
streetcoder123, line 53: That's nice, we have access to the coverage!

src/client/src/App.test.tsx
```diff
@@ -1,9 +1,18 @@
...
-import * as ReactDOM from "react-dom";
+import { shallow } from "enzyme";
+
 import App from "./App";
 
-it("renders without crashing", () => {
-  const div = document.createElement("div");
-  ReactDOM.render(<App />, div);
-  ReactDOM.unmountComponentAtNode(div);
+describe("App", () => {
```
streetcoder123, line 6: You might want to be more descriptive in the describe statement. I see that for now, we are not testing it a lot but I think a description like "App Component Renders" would be more intuitive.

src/client/src/components/Card/Card.test.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from "react";
+import { shallow } from "enzyme";
+import { Card } from "./index";
+
+describe("Card", () => {
```
streetcoder123, line 5: Similar thing here. You can probably write the description as "Card Component renders appropriately". In fact, in future, when we add more tests we can be more precise on what are we trying to test with this describe statement.

src/client/src/setupTests.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * This setup file configures the Enzyme Adapter and is executed before running the tests
+ * https://facebook.github.io/create-react-app/docs/running-tests#initializing-test-environment
+ */
+configure({ adapter: new Adapter() });
+
+const intlProvider = new IntlProvider({ locale: "en" }, {});
+const { intl } = intlProvider.getChildContext();
+global.intl = intl;
```
streetcoder123, line 13: Just for curiosity, how does integrating 'intl' in test configuration work/help?

streetcoder123 (commented):
Left some comments. They are minor. Overall, it looks good!

src/client/src/App.test.tsx
jsondoo: that's done in the `it` statements

src/client/src/setupTests.ts
jsondoo: Some components use `intl` prop, I think it acts as a mock prop. You can try console.log and see the contents. @piotrmark maybe you can help clarify?

src/client/src/App.test.tsx
streetcoder123: I agree, your `it` statement says that `renders without crashing`.
There can be more `it` statements within the `describe` statement that say for instance: `renders a button` or `crashes for a particular prop`. I feel we should have a more descriptive `describe` statement. This would really help as we would be developing more tests.

jsondoo: Okay sure, different examples did it differently but i'm open to this

src/client/package.json
jsondoo: Yea, it's kinda misleading though since its already at 17% lol

src/client/src/setupTests.ts
jsondoo: just to be clear, this was for handling components that use `react-intl`

https://github.com/formatjs/react-intl/blob/master/docs/Testing-with-React-Intl.md

src/client/src/App.test.tsx
piotrmark: IMO it's ok. I usually see describe("Tested component", ...)

src/client/src/setupTests.ts
piotrmark: Yup. As @jsondoo said, thanks to that you can just set intl: global.intl in your tests.

piotrmark (approved):
Good job!

src/client/src/components/Card/Card.test.tsx
piotrmark: Like above. IMO it's good :)

src/client/src/App.test.tsx
streetcoder123: Sure :)

streetcoder123 (approved):
Nice work :)

jasmingoh (approved):

PR #695, opened 2019-06-06
Angular all folders name convention

### LINKED ISSUE #668: Angular naming conventions should be respected
I guess this issue requires a lot of minor changes, but it is very important since the goal of WebTS is to generate code that follows _best patterns and practices_. Since the goal here is to support Angular, the scaffolded code should follow [Angular's Style Guide](https://angular.io/guide/styleguide). Following is a non-exhaustive list of changes to be made in the current code:

- [x] Output() properties should not be suffixed with `EventEmitter`, nor should they be prefixed with `on`. [Reference](https://angular.io/guide/styleguide#avoid-aliasing-inputs-and-outputs)
- [x] Folders & filenames should all be [kebab-case only](https://angular.io/guide/styleguide#symbols-and-file-names).
- [ ] Router modules should be separated from their modules, and the main routing-module should reflect all that's possible within the app. Currently, when looking up the AppModule's routing, it's impossible to know that there are other routes in the app. (_Modules should also be lazy-loaded, since it isn't much work added and shows best practices in terms of bundling size_)
- [x] Properties should not be prefixed with underscores `_`. 
i.e. **ListFormComponent** has `_textField`, which is even more odd since it's actually public by default
- [x] Useless getters/setters should be avoided (_ListFormComponent_)
- [x] Assets should always be included from the `assets/` folder (svg images as well)

If any other Angular Developers would like to hop in and fill in other critical styling issues, we could address all of the related issues in this thread.

One **critical point** is that, in its current state, the generated code **does not even pass** its own generated linting file. I think this should be addressed quickly, as it does not give out a good first glance about the quality of what's being generated.

Tanya0609:
Github Issue: #668, #681
This change is to address one of the suggestions made by angular exeperienced users and experts: 
1. "all folders and symbol names should be in **lower kebab case**, eg, Blank --> blank, Master_Details.module.ts --> master-details.moduke.ts" 
2. Output() properties should not be suffixed with EventEmitter

closes #668

**How to test:**
1. generate an angular project with all 4 type pages selected
2. open the project in vs code and check the folder names 
* there is an exception for master_detail folder name and master_detail.module.ts file, these two can not be addressed at the moment cause it has to do with CoreTemplateStudio.

**Expected Result:** 
name-convention


jsondoo (commented):
agul

small (?) coupling issue, this also makes the pages cards lowercase

Tanya0609:
Good Catch! @jsondoo The **name** attribute under "templates/Web/Pages/Angular.[Page_Name]/.template.config/template.json" file is coupled with **folder name generated**, **page name showed in the card**, **name in generated nav bar** and **module name generated**. 

If I revert the change in the name attribute right now, the folder name will be inconsistent and violates Angular best practice. I would like to address the name convention first, then address the decoupling 
in a soon future PR. 

To resolve decoupling, it will be a big change in generation experience, which will affect and require change for React and Vue generation experience too.

SahilTara:
couldn't people still rename things to "Grid Hello" or something? @Tanya0609 Am I missing the point here? Grid Hello would become Grid_Hello.

Tanya0609:
My observation so far is that, let us say "name = blank " in templates/Web/Pages/Angular.Blank/.template.config/template.json. The name attribute is used to replace the variable name AngularBlank (a folder name) and AngularBlank.module.ts (a file name). And this name also showed in the page card, so when user selects a blank page, a default name "blank" is generated. Correct me if I am wrong. I did not think about renaming case cause I thought we don't allow people to do so.

SahilTara:
Yeah that is what will be used if no one changes the name, notice the things are all input boxes, try typing a custom name in for the page, notice the generated output retains the name with a slight modification of "_" added.

Tanya0609:
closed the PR since it could not address naming convention totally at the moment. Will reopen one once it gets resolved

PR #705, opened 2019-06-11
Create Base Templates for Flask

### LINKED ISSUE #704: Create Base Templates For Flask
We need to make flask templates so that they a minimal usable flask application could be generated with any given front-end framework.

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/27725

This PR adds base templates for a new Framework 'Flask' to the application (in preview mode).

**How to Check**
- Using the wizard, try to generate a Flask Application with any of the available front-end frameworks.
- The wizard should successfully generate the project for you.
- Go to the project and run: `yarn install` to install the dependencies
- Run `yarn start` to start the project
- Currently, only the blank page would work.
- Support for other pages would be added using composition templates in the coming PRs.

closes #704


templates/Web/Pages/React.List/.template.config/template.json
```diff
@@ -12,7 +12,7 @@
     "wts.platform": "Web",
     "wts.projecttype": "FullStackWebApp",
     "wts.frontendframework": "ReactJS",
-    "wts.backendframework": "NodeJS",
+    "wts.backendframework": "all",
```
piotrmark, line 15: Do we need this change here as it works for blank page only and this is React List?

piotrmark (approved):
Looks good

templates/Web/Pages/React.List/.template.config/template.json
streetcoder123: We need this so that React can generate a list template for all backend frameworks and not just NodeJS. Otherwise, if you select a list framework with a Flask application, it gives you an error that templates cannot be generated.

annielovesjs (approved):
it works!

Tanya0609 (approved):
generated flask with React and Angular, both works. also generated node project, it doesn't break anything. approve

PR #711, opened 2019-06-11
some naming convention, file structure, and routing

### LINKED ISSUE #668: Angular naming conventions should be respected

Tanya0609:
This PR is to addressed all the remains changes regarding to improve Angular code quality based on Angular experts suggestion ( mainly from #668). The main test point for this PR, is the generated project works and generated code follows name convention. 

closes #668

**Summary:** 
1. naming conventions
- nav-bar
- footer
- grid subfolders
- master detail subfolders
- list subfolders
(some folders still camel case or has _, this will be fixed through change in CoreTS, not in this PR)

2. fix file structure (component should 4 files)
3. add routing files and lazy loading 
4. @output properties should not be suffixed with EventEmitter (just renaming)

**How to test:**
1. generate an angular project with 4 type of pages selected 
2. start the project, the project should start properly
3. run "npm run lint" or "yarn lint" should only has 2 errors (related to master detail, which will be fixed by change in CoreTS, not in this PR)

**Expected:**
index1

index2


templates/Web/Pages/Angular.MasterDetail/src/app/app-shell/AngularMasterDetail/AngularMasterDetail-routing.module.ts
```diff
@@ -0,0 +1,19 @@
+import { NgModule } from '@angular/core';
+import { Routes, RouterModule } from '@angular/router';
+
+import { MasterDetailComponent } from './master-detail.component';
+
```
markAtMicrosoft, line 5: Remove one line

templates/Web/_composition/AngularJS/Page.Angular.AddRoute/src/app/app-routing.module_postaction.ts
```diff
@@ -0,0 +1,22 @@
+﻿import { NgModule } from '@angular/core';
+import { Routes, RouterModule } from '@angular/router';
+
+const routes: Routes = [
+//^^
```
markAtMicrosoft, line 5: Don't check-in comments if you can avoid it.

markAtMicrosoft (changes requested):

templates/Web/_composition/AngularJS/Page.Angular.AddRoute/src/app/app-routing.module_postaction.ts
SahilTara: That is actually a macro used by the corets engine. It actually translates to "Put the following code (enclosed with //{[{ and //}]}) closer to the context given near the "//}]}" in this case it would insert the code closer to the ```@NgModule``` rather than the line "const routes: Routes". 

In other words it inserts the route at the bottom of the list rather than the top.

Tanya0609: As sahil mentioned, this is the format needed for CoreTS to merge two files together, not a regular comments

templates/Web/Pages/Angular.MasterDetail/src/app/app-shell/AngularMasterDetail/AngularMasterDetail-routing.module.ts
Tanya0609: The reason there is line to seperate import built in package and import custom packages (the same way Angular CLI does it). I have this format for every module file.

templates/Web/_composition/AngularJS/Page.Angular.AddRoute/src/app/app-routing.module_postaction.ts
markAtMicrosoft: Sorry, my lack of knowledge in this code.

markAtMicrosoft (approved):

streetcoder123 (approved):

piotrmark (approved):

MehaKaushik (approved):

PR #712, opened 2019-06-12
Abstracts CoreTS interface in extension

jcoc611-microsoft:
## Changes

Abstracts away the CoreTS interface in the extension to a single class that can interact with the templates. The reasoning behind this change is two-fold. First, we are going to be changing the way the extension talks to the CoreTS generator soon as the server is too heavy of an implementation. Second, we are working on generating web apps from templates in the pipeline, and need to reuse the functionality outside the extension.

Aside from abstracting the interface, this PR has some goodies like reusing connections for SignalR WebSocket connections. It would be nice if the interns from previous cohort @SahilTara and @trevorNgo could verify this doesn't break any edge cases you found.

This work is missing updating the client code to use the CoreTemplateStudio class to get the list of frameworks, etc. It should be done by going through the extension via vscode messaging, because otherwise the interface is not transparent to using something different than a HTTP server to communicate with CoreTS.


annielovesjs (dismissed):
Tested it out by going through the generation process for the wizard and hosting the created project and it works as intended

piotrmark (dismissed):

src/extension/src/constants.ts
```diff
@@ -160,7 +160,24 @@ export const CONSTANTS = {
     SIGNALR_API_SYNC_METHOD_NAME: "SyncTemplates",
     SIGNALR_API_GENERATE_METHOD_NAME: "Generate",
     MAX_SYNC_REQUEST_ATTEMPTS: 51,
-    SYNC_RETRY_WAIT_TIME: 250
+    SYNC_RETRY_WAIT_TIME: 250,
+    Endpoints: {
```
Tanya0609: If it is constant, should it be ENDPOINTS?

jcoc611-microsoft: Good question! I copied it over from the [client constants](https://github.com/microsoft/WebTemplateStudio/blob/dev/src/client/src/services/constants.tsx) but I could update it to uppercase

piotrmark (approved):

annielovesjs (approved):

PR #727, opened 2019-06-13
angular remove constant file and put content in environment files

Tanya0609:
This PR is to address Angular expert's suggestion on constant file in #681 
Task:https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/WTS%20Team/Intern%20GitHub/Y19-VS/04

**Summary:**
" 1. Move these constants (endpoints) to src/environments/environment.*.ts` (you will have to duplicate them)
2. Don't use constants for error messages and just hard code the messages in their components (giving the fact that the rest of your text and description are in plain English)."

**How to test:**
1. generate an angular project with 4 types of pages selected 
2. able to start the project
3. look at the project file structure: 
          - endpoints should be inside both environment files under /environment folder
          - there is no constant.ts file under /src


piotrmark (approved):

streetcoder123:
The generated project generates an environment.ts file which is something like this when all pages are selected:
```
export const environment = {
  production: false,
  endpoint: {
    masterdetail : '/api/masterdetail',
    list : '/api/list',
    grid : '/api/grid',
  }
};
```
This is a JavaScript object but we have a comma after the last attribute of the object. This is not a common thing you would see in a JavaScript object. In fact, some linters even flag this as a lint error.

What do you think about it?

templates/Web/Pages/Angular.Grid/src/app/app-shell/AngularGrid/grid.component.ts
```diff
@@ -1,17 +1,17 @@
...
+
 @Component({
   selector: 'app-grid',
   templateUrl: './grid.component.html',
   styleUrls: ['./grid.component.css']
 })
 export class GridComponent implements OnInit {
 
-  WarningMessageText = CONSTANTS.ERROR_MESSAGE.GRID_GET;
+  WarningMessageText = 'Request to get grid text failed:';
```
streetcoder123, line 14: I don't know if I am missing something but why is it hard coded?

templates/Web/Pages/Angular.Grid/src/app/app-shell/AngularGrid/grid.component.ts
```diff
@@ -38,7 +38,7 @@ export class GridComponent implements OnInit {
       },
       error => {
         this.WarningMessageOpen = true;
-        this.WarningMessageText = `${CONSTANTS.ERROR_MESSAGE.GRID_GET} ${error}`;
+        this.WarningMessageText = `Request to get grid text failed: ${error}`;
```
streetcoder123, line 41: Same thing here

piotrmark:
> The generated project generates an environment.ts file which is something like this when all pages are selected:
> 
> ```
> export const environment = {
>   production: false,
>   endpoint: {
>     masterdetail : '/api/masterdetail',
>     list : '/api/list',
>     grid : '/api/grid',
>   }
> };
> ```
> 
> This is a JavaScript object but we have a comma after the last attribute of the object. This is not a common thing you would see in a JavaScript object. In fact, some linters even flag this as a lint error.
> 
> What do you think about it?

Trailing comas in ES6 is a common practice (we even require it in our project).

NIT: The property could be named `endpoints`

templates/Web/Pages/Angular.Grid/src/app/app-shell/AngularGrid/grid.component.ts
piotrmark: I believe it's because 

> Don't use constants for error messages and just hard code the messages in their components (giving the fact that the rest of your text and description are in plain English).

The best way to handle it would be by localization (https://angular.io/guide/i18n), but as we're not doing this, I agree that constants don't bring any value because these strings are not being reused.

streetcoder123:
@piotrmark Oh, I did not know about that. I thought that we should not have trailing commas. Anyways, that is good to know :)

streetcoder123 (approved):

jsondoo:
It's not a big deal, but Prettier actually gets rid of trailing commas, probably okay either way tho

jsondoo (approved):

piotrmark:
> It's not a big deal, but Prettier actually gets rid of trailing commas, probably okay either way tho

Actually, you can set this up how you like, but yeah it's not super important ;)
https://prettier.io/docs/en/options.html#trailing-commas

PR #730, opened 2019-06-13
angular folder name should be kebab case

Tanya0609:
task:https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/WTS%20Team/Intern%20GitHub/Y19-VS/04

This PR is to achieve: **folder name should be kebab case and module name should be pascal case.** These two names were coupled with whatever the default name or user input name, also "-" will be converted into "_". Now with Sibille's change, I am able to get different casing names from dedault name or user input name. 

Sibille stored cased names in variable "Param_SourceName_Kebab" and "Param_SourceName_Parscal", so I just need to query them in template.json file and use it in the correct places, i.e. folder name, module name.

**How to test:**
1. generate a project with 4 types of page selected
2. the app should start properly
3. do "yarn lint" or "npm run lint", all files should pass

**Expected:**
fodler-name


jsondoo (commented):
image

Footer is still capitalized

jsondoo (commented):
verified `yarn lint` passes with various cases 👍

Tanya0609:
Footer should have been addressed in a different PR that has been merged. So this happened either because this PR originally don't contain the change from the other PR (which should not be a problem anymore since I just rebased it with the newest dev branch) OR this is something to do with GitHub, it won't detect the two same words one starts with lowercase and one starts with uppercase, so that it did not update your code in your VS code. So that your code still have Footer instead of footer as a folder name.

jasmingoh (approved):
Works for me! I generated a project using Angular as the frontend framework, and all the generated folder names are in kebab case. 

Tried naming some pages with consecutive dashes/underscore, and it seems like those name are cleaned up (e.g. Grid___test ---test2 becomes grid-test-test2)

jsondoo (approved):

piotrmark (approved):

PR #734, opened 2019-06-14
Update submodule to allow adding casing for params

sibille:
This allows to add casings for parameters, for more info on config see:
https://github.com/microsoft/CoreTemplateStudio/issues/144

edited by Tanya:
**How to test:**
1. generate a angular project with at least two pages
2. open the project in vscode and nav to "app-routing.module.ts" file
3. you should see as home url is kebab case (like the following)
home-url


Tanya0609:
I was trying to test your change. What I did was pull your PR (cause this PR #734 is on your forked project) and then tried to use the "Param_HomePageName_Kebab" in home url. Then I just start the extension and generate a project with more than 1 pages. But it seems not working, cause generated home url ended up being just "Param_HomePageName_Kebab".

geerated

Did I miss anything so that your change not actually in my local? Or could you put your change directly under this repo like your pervious PR?

This was the command I was using to pull:
git fetch origin pull/734/head:734 && git checkout pr/734 --recurse-submodules

sibille:
@tanya0609, this should work fine from the fork too, but there was a bug when using parameter casing on a composition template only. I've fixed it and updated the PR, this should work now.

Tanya0609 (dismissed):
It works now! Thank you @sibille.

streetcoder123 (approved):
This works with generated Angular Application

piotrmark (approved):

jsondoo (approved):

Issue #737, opened 2019-06-14
Refactor Flask Endpoints into separate template to make generation smoother


streetcoder123:
We need to refactor a few things:
- Add Grid Merging to separate the logic of adding grid endpoints from adding grid routes
- Add Another template 'AddEndPoints' that runs only if at least one page is selected out of list, grid and master detail so that we need to create endpoints constants for them.

PR #738, opened 2019-06-14
Refactor Flask Endpoints into separate template to make generation smoother

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/28692

**Changes**
- Add Grid Merging to separate the logic of adding grid endpoints from adding grid routes
- Add Another template 'AddEndPoints' that runs only if at least one page is selected out of list, grid and master detail so that we need to create endpoints constants for them.

c;loses #737


jasmingoh (approved):
Code looks good and tested the changes out by creating Flask projects with and without grid page.

annielovesjs (approved):
works properly when tested :)

jcoc611-microsoft (approved):

PR #741, opened 2019-06-14
Remove project type

jsondoo:
This changes the UI to remove rendering components related to 'Project Type'
We decided to keep the actual components like `SelectWebApp` in case we decide to revert this later and add more types later.

Please test for any breaking changes that I may have missed!


src/client/src/components/LeftSidebar/index.tsx
```diff
@@ -22,25 +18,21 @@ const messages = defineMessages({
...
     defaultMessage: "1. New Project"
   },
-  projectType: {
-    id: "leftSidebar.projectType",
-    defaultMessage: "2. Project Type"
-  },
   frameworks: {
     id: "leftSidebar.frameworks",
-    defaultMessage: "3. Frameworks"
+    defaultMessage: "2. Frameworks"
```
jcoc611-microsoft, line 23: Would it make more sense to have these numbers be auto generated based on page order? 

```javascript
`${index}. ${defaultMessage}` 
```

Edit: Numbers must also be internationalized

jcoc611-microsoft (dismissed):
Tested and client still works as expected, project type is no longer present

src/client/src/components/LeftSidebar/index.tsx
jsondoo: Yea it would make more sense - oo I did not know numbers need to be i18n'd

jsondoo:
Made a small change to make sure the Redux state matches up exactly

https://www.diffchecker.com/zDTIMOtg
I generated a React + Node app on dev and my branch, and then compared the Redux tree. 
(left side is dev, right side is after my changes)

src/client/src/containers/ReviewAndGenerate/index.tsx
```diff
@@ -123,7 +107,6 @@ const mapDispatchToProps = (
 });
 
 const mapStateToProps = (state: AppState): IStateProps => ({
-  projectTypeRows: WizardSelectors.getProjectTypeRowItemSelector(state),
```
Tanya0609, line 126: Since this selector won't be used during our scope, maybe it will be good in the WizardSelector.ts to make a comment or comment out the section of code related to this selector, but this is just a suggestion

src/client/src/reducers/wizardSelectionReducers/selectWebAppReducer.ts
```diff
@@ -13,8 +13,8 @@ import WizardInfoType from "../../actions/wizardInfoActions/wizardInfoActionType
 const initialState = {
   title: "Fullstack Web Application",
   internalName: "FullStackWebApp",
-  version: "",
-  author: ""
+  version: "v1.0",
+  licenses: ""
```
Tanya0609, line 17: what does this change for?

src/client/src/components/LeftSidebar/index.tsx
jcoc611-microsoft: I think most notably right-to-left languages are going to be different. But also, other languages have different [number symbols](https://en.wikipedia.org/wiki/Hindu%E2%80%93Arabic_numeral_system#Symbols) too, even if 0-9 symbols are understood in a modern context in a given locality.

src/client/src/containers/ReviewAndGenerate/index.tsx
jsondoo: I don't think this is a big deal since I decided to remove the minimum in case we decide to reuse components or selectors back in later. I could also make a PR where we completely clean up everything later once we decide

src/client/src/reducers/wizardSelectionReducers/selectWebAppReducer.ts
jsondoo: honestly, I'm not 100% sure what these attributes are for, but I made the initial state to match what the state should look like after selecting a full stack web app

jsondoo: looking at `generationSelector.ts` we only use the internalName attribute so I dont think version/licenses are even used right now

annielovesjs (approved):

src/client/src/components/LeftSidebar/index.tsx
piotrmark: Yeah, `${index}. ${defaultMessage} ` wouldn't work for rtl languages, so maybe it's better to leave as-is and localize the whole thing?

piotrmark (approved):

Tanya0609 (approved):

src/client/src/components/LeftSidebar/index.tsx
jsondoo: I will leave this as is

PR #743, opened 2019-06-14
angular fix ng build prod errors

Tanya0609:
**How to test:**
1. generate a angular project with 4 types of page selected 
2. do npm install
3. do ng build --prod, there should be no errors 

To address one suggestion from Angular Expert #681 

Errors to be addressed by this PR:
ngbuild


jsondoo (approved):
Tried both with and without `--prod` and it looks good

piotrmark (approved):

streetcoder123 (commented):
@Tanya0609 Your change works. I think that we should add `ng build --prod` as a script to package.json file. This would allow users who do not have `ng` installed globally on their machine would also be able to run the script and check for themselves.

Tanya0609:
Can incorporate this change into a different PR

streetcoder123 (approved):

PR #758, opened 2019-06-18
Dependency checker for frameworks page

jsondoo:
**How it works**
- When the Select Frameworks page loads, the component uses `vscode.postMessage` to tell the extension to figure out the dependency installation state
- Extension receives the command, calls `checkDependency` function in dependencyChecker.ts
- Extension returns a payload with the installation state to the client
- Client listens and receives the payload in `App.tsx` in the `case EXTENSION_COMMANDS.GET_DEPENDENCY_INFO` and dispatches an action to update the Redux store with the dependency info
- `SelectableCard` uses the information from the Redux store and renders a `DependencyInfo` component

**Client side changes**
- App.tsx: add a new listener that listens for the payload from the extension
- Added new type, action and reducer with WIZARD_INFO_TYPEKEYS.UPDATE_DEPENDENCY_INFO
- Yellow warning svg
- DependencyInfo component: takes in framework name and installation state and displays a warning (this is gated with a preview flag)
- SelectFrameworks is moved into containers folder from components since I use the Redux store, namely the vscode API to send payloads to the extension when this component is loaded (frameworks page)
- New prop `isFrameworkSelection` passed down to `SelectableCards` because this component is actually shared between all the pages, I only want the dependency info showing on the frameworks page
- 2 new strings in translations file
- mockVsCodeApi - this is used when the client is run in the browser and mimics communication with the extension

**Extension side changes**
- src/constants.ts: Added new command and extension module for dependency checker
- src/controller.ts: Added dependency checker to the extensionModuleMap
- src/utils/dependencyChecker.ts: logic for checking dependency (we are currently checking for node and python3.x, the python command depends on users' operating system)

**How to test**
- Build the project
- Go to VS Code setting and enable preview mode
- Launch the extension and go to the Framework page
- Try installing/uninstalling Python/Node
- Go to another page, and come back to the Framework page, and the dependency checker status should change
- DependencyInfo should not show up if preview mode is not enabled


jsondoo:
Expected behavior 
expected

Also apologies for the big PR, we weren't sure if we should split it up a bit, PR for extension code and then PR for client side + communication, but we thought combining it would be easier to test

src/client/src/actions/wizardInfoActions/updateDependencyInfo.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { updateDependencyInfoAction } from "./updateDependencyInfo";
+import { WIZARD_INFO_TYPEKEYS } from "./typeKeys";
+
+describe("updateDependencyInfo actions", () => {
+  it("should create an action", () => {
+    expect(updateDependencyInfoAction({ a: 1 })).toEqual({
```
piotrmark: NIT: can we declare this object above and use it by reference?

src/client/src/components/DependencyInfo/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+    let dependencyMessage: string = installed
+      ? intl.formatMessage(messages.installed)
+      : intl.formatMessage(messages.notInstalled);
+
+    let icon: any = installed
+      ? getSvg.getGreenCheckSvg()
+      : getSvg.getWarningSvg();
+
+    // map framework name to dependency name
+    let dependencyName = frameworkName === "Flask" ? "Python" : "Node";
```
piotrmark, line 52: Can we declare a map instead and then use let's say `let dependencyName = frameworkToDependencyMap[frameworkName]`?

src/client/src/components/DependencyInfo/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    let downloadLink: string = "";
+    if (dependencyName === "Node") {
+      downloadLink = "https://nodejs.org/en/download/";
+    } else if (dependencyName === "Python") {
+      downloadLink = "https://www.python.org/downloads/";
+    }
```
piotrmark, line 59: Same here, a map would be more clear.
Or maybe even object like:
`"NodeJS": {
    "dependencyName": "Node",
    "downloadLink": "https://nodejs.org/en/download/"
},
"Flask": {...}`

src/client/src/components/DependencyInfo/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+      <a
+        target={"_blank"}
+        href={downloadLink}
+        className={classnames(styles.dependencyContainer, {
+          [styles.disabled]: installed,
+          [styles.borderGreen]: installed,
+          [styles.borderYellow]: !installed
+        })}
+      >
+        <img className={styles.icon} src={icon} alt="" />
```
piotrmark: This image has a meaning, so for accessibility reasons, alt tag should not be empty. You can provide image description based on the same logic you provide the icon.

piotrmark: + it should be a localized string

src/client/src/containers/SelectFrameworks/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+      }
+    });
+    vscode.postMessage({
+      module: "DependencyChecker",
+      command: "check-dependency",
+      payload: {
+        dependency: "python"
+      }
+    });
+    console.log("Posting check-dependency to extension");
```
piotrmark: Do we log every postMessage? If not and this was added for debugging purposes, it should be removed :)

src/client/src/containers/SelectFrameworks/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { getVSCodeApiSelector } from "../../selectors/vscodeApiSelector";
+
+import DependencyInfo from "../../components/DependencyInfo";
+
+import { AppState } from "../../reducers";
+
+interface ISelectFrameworksProps {
+  vscode: any;
+  dependencies: any;
+  selectedBackend: any;
```
piotrmark: Are these 2 props used somewhere?

src/client/src/containers/SelectFrameworks/index.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+type Props = ISelectFrameworksProps;
+
+class SelectFrameworks extends React.Component<Props> {
+  componentDidMount() {
```
piotrmark, line 23: It would be nice to perform this only in preview mode, because now only display logic is behind the flag, but commands will be executed anyway and the commands are most scary here :)

src/client/src/reducers/dependencyInfoReducers/index.spec.ts
```diff
@@ -0,0 +1,36 @@
...
+    const expectedState = {
+      node: {
+        installed: true
+      }
+    };
+    expect(dependencyInfo({ node: { installed: false } }, mockAction)).toEqual(
+      expectedState
+    );
+  });
+});
```
piotrmark, line 36: Thank you for beautiful unit tests!

Tanya0609:
I have both Node and Python in my local. The node dependency works well (says NodeJS is detected) for React, Angular, Node and Vue.

(I am on your branch, but I still have select project type page in my extension, but in your attached video, you don't have it)

When I click on Flask framework, all the display on the extension just gone.
Screen Shot 2019-06-18 at 8 14 21 PM

Screen Shot 2019-06-18 at 8 13 49 PM

src/extension/src/utils/dependencyChecker.ts
```diff
@@ -0,0 +1,75 @@
...
+            state = await this.checkForPython3();
+        } else {
+          state = true;
+        }
+      } else if (stderr.length > 0) {
+        state = false;
+      } 
+    } catch (err) {
+      if (userOnWin) {
+        state = await this.checkForPython3();
```
piotrmark: Not sure if this logic is right.
Let's say payload.dependency is "node", userOnWin is true - why we check for Ptyhon3 here?

In gereral i think this function will benefit having 2 separate paths for python and node

src/client/src/components/DependencyInfo/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+    defaultMessage: " not detected. Click to install."
+  }
+});
+
+/*
+ * Props:
+ * - frameworkName: "NodeJS" | "Flask"
+ * - installed: boolean, true if correct version installed, otherwise false
+ */
+class DependencyInfo extends React.Component<Props> {
```
piotrmark, line 35: This is a new component, so a great candidate for unit tests as well :)

piotrmark (changes requested):

piotrmark:
> I have both Node and Python in my local. The node dependency works well (says NodeJS is detected) for React, Angular, Node and Vue.
> 
> (I am on your branch, but I still have select project type page in my extension, but in your attached video, you don't have it)
> 
> When I click on Flask framework, all the display on the extension just gone.
> Screen Shot 2019-06-18 at 8 14 21 PM
> 
> Screen Shot 2019-06-18 at 8 13 49 PM

That's why it's better to have all the logic behind preview flag ;)

src/extension/src/utils/dependencyChecker.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  async checkDependency(message: any): Promise<IPayloadResponse> {
+    var userOS = os.platform();
+    var userOnWin = userOS.indexOf('win') === 0;
+    var name = message.payload.dependency === 'python' && !userOnWin ? 'python3'
+                                                                     : message.payload.dependency; 
+    var state;
+    try {
+      const { stdout, stderr } = await exec(name + ' --version');
```
piotrmark: Security concern: should we sanitize the name?
What if someone posts malicious command?
At the other hand, if the message cannot come from outside of extension client, maybe it's not an issue.

@jcoc611-microsoft , thoughts?

jasmingoh: How do we sanitize the name? 

This part of the code can only be triggered through the client's posted messages, so no messages can come from outside of the client.

src/extension/src/utils/dependencyChecker.ts
jasmingoh: My bad, I meant to check for if the name is 'python' here, because the command stays as python if the user is on Windows. I guess I generalized too much here.

The reason I did not separate them into 2 different path is because we have 4 cases in total: (1) check for node on unix (2) check for node on windows (3) check for python 3 on unix (4) check for python 3 on windows, and it is only the last one that is harder to handle. 

I totally see your point of separating python and node. Thanks for the suggestions! Will make the changes :) (Just realize it, if we have 2 path for node and python, we won't have to worry about user inserting malicious commands too)

src/client/src/reducers/dependencyInfoReducers/index.spec.ts
jsondoo: You're welcome :)

src/extension/src/utils/dependencyChecker.ts
jsondoo: @jasmingoh do you want to elaborate on this?

src/client/src/components/DependencyInfo/index.tsx
jsondoo: I agree, I was planning on refactoring it later since we only have two cases right now, but I could do it in this PR if you want

src/client/src/containers/SelectOption/index.tsx
```diff
@@ -223,6 +224,7 @@ class SelectOption extends React.Component<Props, ISelectOptionState> {
             return (
               <SelectableCard
                 key={`${cardNumber} ${title}`}
+                isFrameworkSelection={this.props.isFrameworkSelection}
```
SahilTara: nit can we destructure this like we do with everything in line 223?

src/extension/src/utils/dependencyChecker.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  private defineCommandMap(): Map<
+    ExtensionCommand,
+    (message: any) => Promise<IPayloadResponse>
+  > {
+    return new Map([[ExtensionCommand.CheckDependency, this.checkDependency]]);
+  }
+
+  private async checkForPython3() {
+    var state;
```
SahilTara: nit please use let here, maybe even define the type? 
```ts 
let state: boolean
```

src/extension/src/utils/dependencyChecker.ts
```diff
@@ -0,0 +1,75 @@
...
+    }
+    return state;
+  }
+
+  private outputIsPython27(stdout: any, stderr: any) {
+    return stdout.indexOf(PYTHON27) === 0 || stderr.indexOf(PYTHON27) === 0;
+  }
+
+  async checkDependency(message: any): Promise<IPayloadResponse> {
+    var userOS = os.platform();
```
SahilTara: Same as above can we use let here?

src/client/src/containers/SelectFrameworks/index.tsx
jsondoo: removed!

src/client/src/containers/SelectFrameworks/index.tsx
jsondoo: no, good catch

src/client/src/containers/SelectOption/index.tsx
jsondoo: done

jsondoo:
@Tanya0609 it seems like this happens on Mac/Linux, was this happening when you used Windows as well? i will try to debug this once I set up the other OS

Edit: figured out why, some miscommunication happening with the payload between extension and client with python3, should work cross platform now

src/client/src/containers/SelectFrameworks/index.tsx
jsondoo: makes sense, the postmessage is in preview now

jasmingoh:
Just refactored extension side dependency checker code as per @njindal239 & @SahilTara 's suggestion. 

We are running the `python3` and `python` command regardless of which OS the user is on. This covers the case if user sets a python alias for Python 3.X on Unix (we were only running `python3` command on Unix).

We are only running the python launcher command, `py -3`, for Windows because Python Launcher is installed by default when user installs Python 3.X. Another reason for including the Python Launcher command is because `python --version` and `python3 --version` are inconsistent when user have both Python 2.X and/or Python 3.X installed on Windows, and `py -3 --version` is more reliable.

The only case that we cannot cover is when user is on Windows, has Python2.X and Python3.X installed, but no Python Launcher installed. The dependency checker would return a false negative because none of the command would work. `python --version` would return Python 2.X, `python3 --version` would return blank and `py -3 --version` will return an error.

image
(But this is very unlikely to happen, as mentioned above, Python Launcher comes with Python 3.X, and the user has to also have Python 2.X installed for the dependency checker to return a false negative)

SahilTara:
Just curious @jsondoo do we care about the minimum installed version. Since our templates were built against node version 10.x.y (forget the exact version) so if for some reason we have someone using node 8.x.y or 7.x.y for some reason would we run into any issues? Same with python, we built against 3.7.3, will we have any problems with people having python 3.2.0 or something?

By the way small nit in terms of UI shouldn't the check marks be on the same side for both selected and the checkmark in "Nodejs detected"? Notice the right vs left.

xtinah-w:
@SahilTara Speaking on behalf of the UI for the cards, we actually have a new card design coming out based on user research we've conducted (a more concise description, fixing the highlight on the cards, including version number), so this card UI won't be remaining the same. For dependency checker, it makes more sense for the symbol to be on the left.

During our user testing sessions, people relied on the highlight to show them which card they had selected (thus, we're fixing the orange highlight) which caused confusion and anxiety. 

Traditionally, most warnings have a symbol on the left with the warning text on the right. It's rare to see an icon on the right side. 
warning

jasmingoh:
@SahilTara I've tested the generated Flask application with Python 3.6, and it works fine. We will just change the supported version to Python 3.6.X or Python 3.7.X (they are the most commonly used Python 3.X versions)

We will test for node as well.

SahilTara:
My question is more so what if the user has python 3.5 installed currently. The current regex will allow it do we want to allow them? If so all is good, if not should there be feedback telling the user they have below the minimum supported version? @jasmingoh

SahilTara:
@xtinah-w sounds good yeah it did make sense that the dependency thing was on the left just was wondering if the selected check should also be on the left. But since the ui is going to change will you be removing that check (the one beside the selected) altogether?

xtinah-w:
@SahilTara Truthfully, when I was testing and creating a new prototype, I didn't use the "Selected" label and no one commented about any doubts that it was selected since the big highlight was a visual indicator that it was chosen. We may just keep in the "Selected" for accessibility, but it's something we're evaluating right now. 

Thanks for the concern and I'll keep it in mind for the future design!

jsondoo:
@SahilTara good point, initially we were going to support python 3.6 and 3.7 for Python, but @jasmingoh is  investigating looking into other versions. For node, I will take a deeper look and see if there's a minimum version that we can't support

src/extension/src/utils/dependencyChecker.ts
```diff
@@ -0,0 +1,84 @@
+import { WizardServant, IPayloadResponse } from "../wizardServant";
+import { ExtensionCommand } from "../constants";
+const os = require('os');
+const util = require('util');
+const exec = util.promisify(require('child_process').exec);
+
+const NODE = 'node';
```
jcoc611-microsoft: Should move these to constants.ts

src/extension/src/utils/dependencyChecker.ts
jcoc611-microsoft: Seems alright for now, as it can only be one of the PYTHON constants

src/client/src/containers/SelectFrameworks/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+    const { vscode, isPreview } = this.props;
+    if (isPreview) {
+      // send messages to extension to check dependency info when this component loads
+      vscode.postMessage({
+        module: "DependencyChecker",
```
jcoc611-microsoft: I think you should use `EXTENSION_MODULES` and `EXTENSION_COMMANDS` for consistency

src/client/src/containers/SelectableCard/index.tsx
```diff
@@ -85,6 +95,20 @@ const SelectableCard = ({
             <CardTitle title={title} />
           </div>
         </div>
+        {isPreview &&
+          isFrameworkSelection &&
+          selected &&
+          (option.internalName === "Flask" ? (
+            <DependencyInfo
```
jcoc611-microsoft, line 102: If you are refactoring the other stuff per @piotrmark you could consider something like this here:

```
 <DependencyInfo
    frameworkName={option.internalName}
    installed={dependencies[option.internalName].installed}
     />
```

src/client/src/mockData/mockVsCodeApi.ts
```diff
@@ -24,6 +24,28 @@ const mockVsCodeApi = () => ({
         case "alert":
           console.log("Command: ", message.alert);
           break;
+        case "check-dependency":
+          window.postMessage(
+            {
+              command: "check-dependency",
```
jcoc611-microsoft: Constants here as well

jsondoo: done

src/client/src/containers/SelectFrameworks/index.tsx
jsondoo: good catch

src/client/src/components/DependencyInfo/index.tsx
jsondoo: okay, unit test will be added after I do the refactoring

src/client/src/containers/SelectableCard/index.tsx
jsondoo: > Can we declare a map instead and then use let's say `let dependencyName = frameworkToDependencyMap[frameworkName]`?

> Same here, a map would be more clear.
> Or maybe even object like:
> `"NodeJS": { "dependencyName": "Node", "downloadLink": "https://nodejs.org/en/download/" }, "Flask": {...}`

(moving @piotrmark comment here since it's hidden)

I will have think a bit more about how this will be refactored (if i want to touch redux store/payloads since i can't directly call as it doesn't have framework names as an attribute, or just include a map from framework -> dependency, or both), but I also don't like how it is right now, maybe a separate PR? And then I will add unit test afterwards

jasmingoh:
Thanks @SahilTara for the qeustion earlier. The minimum version of Python that is required for the generated web Flask application is 3.5, so I've updated the regex to only match 3.5 onwards.

src/client/src/actions/wizardInfoActions/updateDependencyInfo.ts
```diff
@@ -0,0 +1,20 @@
+import { WIZARD_INFO_TYPEKEYS } from "./typeKeys";
+
+export interface IUpdateDependencyInfo {
+  type: WIZARD_INFO_TYPEKEYS.UPDATE_DEPENDENCY_INFO;
+  payload: any;
```
SahilTara: Do you think it would be a good idea if we could had some type safety for this payload?

src/client/src/actions/wizardInfoActions/updateDependencyInfo.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export const updateDependencyInfoAction = (
+  dependencyInfo: any
```
SahilTara: do you think it would be a good idea if we had some type safety for the dependencyInfo?

Possibly something like:
```ts
interface IDependencyInfo {
    dependency: "python" | "node";
    installed: boolean;
}
```

src/client/src/containers/SelectFrameworks/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+import SelectBackEndFramework from "../SelectBackendFramework";
+import SelectFrontEndFramework from "../SelectFrontEndFramework";
+import { getVSCodeApiSelector } from "../../selectors/vscodeApiSelector";
+
+import { AppState } from "../../reducers";
+
+import { EXTENSION_MODULES, EXTENSION_COMMANDS } from "../../utils/constants";
+
+interface ISelectFrameworksProps {
+  vscode: any;
```
SahilTara: Do you think we could get some type safety here too? 

https://github.com/microsoft/WebTemplateStudio/blob/28759f22376ae8b25401fb8591d13dcb7148d168/src/client/src/reducers/vscodeApiReducer.ts#L19-L21

src/client/src/containers/SelectFrameworks/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+    return (
+      <div className={styles.container}>
+        <SelectFrontEndFramework />
+        <SelectBackEndFramework />
+      </div>
+    );
+  }
+}
+
+const mapStateToProps = (state: AppState): any => {
```
SahilTara: Do you think you could define an IStateProps for this?

src/client/src/reducers/dependencyInfoReducers/index.ts
```diff
@@ -0,0 +1,25 @@
+import { combineReducers } from "redux";
+import { WIZARD_INFO_TYPEKEYS } from "../../actions/wizardInfoActions/typeKeys";
+
+export const initialState = {};
+
+export const updateDependencyInfo = (
+  state: any = initialState,
+  action: any
```
SahilTara: Would it be possible to make this slightly more type safe (See how we do it with other reducers). (State + Action).

src/extension/src/utils/dependencyChecker.ts
```diff
@@ -0,0 +1,79 @@
...
+    if (name === CONSTANTS.DEPENDENCY_CHECKER.NODE) {
+      try { 
+        const { stdout } = await exec(CONSTANTS.DEPENDENCY_CHECKER.NODE + ' --version');
+        if (stdout.length > 0) { 
+          state = true;
+        } else {
+          state = false;
+        }
+      } catch (err) {
+        state = false;
```
SahilTara, line 49: Is swallowing the error okay here? Do you think we should be logging this (now or later)?

src/extension/src/utils/dependencyChecker.ts
```diff
@@ -0,0 +1,79 @@
...
+      let userOnWin: boolean = userOS.indexOf('win') === 0;
+
+      if (await this.runPythonVersionCommand(CONSTANTS.DEPENDENCY_CHECKER.PYTHON3)) {
+        state = true;
+      } else if (await this.runPythonVersionCommand(CONSTANTS.DEPENDENCY_CHECKER.PYTHON)) {
+        state = true;
+      } else if (userOnWin && await this.runPythonVersionCommand(CONSTANTS.DEPENDENCY_CHECKER.PYTHON_LAUNCHER)) {
+        state = true;
+      } else {
+          state = false;
```
SahilTara: Nit: Spacing inconsistent (Prettier).

src/extension/src/utils/dependencyChecker.ts
jasmingoh: @SahilTara We could add logging here, but if an error is throw, it means that node/ python is not installed, and since we are only showing 2 states on the client side, I thought swallowing the error is ok. 

Unless we want to give the user more information, then I will be happy to add logging.

SahilTara: Ah okay makes sense. I wasn't sure when the error was thrown so I wanted to ask.

jsondoo:
Updates:
For Python, we detect if they have v3.5+ (LTS at the time of writing is 3.7) 
For Node, we detect if they have v10.15+ (the LTS at the time of writing is 10.16)

src/client/src/containers/SelectableCard/index.tsx
SahilTara: @jsondoo why don't you use a selector or something to map the data into a format you want?

jsondoo: @SahilTara yea, I will do some refactoring in a separate PR

SahilTara: Cool i'm happy with this PR given that you'll be doing the refactorings in a seperate PR. @jsondoo

SahilTara (dismissed):

src/client/src/App.tsx
```diff
@@ -301,6 +307,9 @@ const mapDispatchToProps = (
   updateTemplateGenStatus: (isGenerated: IServiceStatus) => {
     dispatch(updateTemplateGenerationStatusAction(isGenerated));
   },
+  updateDependencyInfo: (dependencyInfo: any) => {
```
SahilTara: Just noticed this, can you add the IDependencyInfo in this PR or your refactoring PR? @jsondoo

src/client/src/App.tsx
```diff
@@ -80,6 +81,7 @@ interface IDispatchProps {
   updateTemplateGenStatusMessage: (status: string) => any;
   updateTemplateGenStatus: (isGenerated: IServiceStatus) => any;
   getVersionsData: (versions: IVersions) => any;
+  updateDependencyInfo: (dependencyInfo: any) => any;
```
SahilTara: Just noticed this as well, can you also add the IDependencyInfo to this in this PR or your refactoring PR? @jsondoo

src/extension/src/utils/dependencyChecker.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  private async runPythonVersionCommand(command: string) {
+    let installed: boolean;
+    try {
+      const { stdout } = await exec(command + ' --version');
+      if (PYTHON3_REGEX.test(stdout)) {
+        installed = true;
+      } else {
+        installed = false;
+      }
```
piotrmark: NIT: could be simplified to `installed = PYTHON3_REGEX.test(stdout)`

src/extension/src/utils/dependencyChecker.ts
```diff
@@ -0,0 +1,80 @@
...
+    let name: string = message.payload.dependency;
+    let state: boolean = false;
+    if (name === CONSTANTS.DEPENDENCY_CHECKER.NODE) {
+      try { 
+        const { stdout } = await exec(CONSTANTS.DEPENDENCY_CHECKER.NODE + ' --version');
+        if (NODE_REGEX.test(stdout)) { 
+          state = true;
+        } else {
+          state = false;
+        }
```
piotrmark: NIT: same here, can be in 1 line :)

src/extension/src/utils/dependencyChecker.ts
```diff
@@ -0,0 +1,80 @@
...
+
+      if (await this.runPythonVersionCommand(CONSTANTS.DEPENDENCY_CHECKER.PYTHON3)) {
+        state = true;
+      } else if (await this.runPythonVersionCommand(CONSTANTS.DEPENDENCY_CHECKER.PYTHON)) {
+        state = true;
+      } else if (userOnWin && await this.runPythonVersionCommand(CONSTANTS.DEPENDENCY_CHECKER.PYTHON_LAUNCHER)) {
+        state = true;
+      } else {
+        state = false;
+      }
```
piotrmark, line 72: NIT: here as well can be shortened to `state = await ... || await ... || (userOnWin && await ...)`, but do it only if YOU find it more readable  :)

piotrmark (dismissed):
Thank you very much for incorporating our feedback, left a couple of minor suggestions, but the code looks really much better now! Good job, can't wait to see it out of preview! :D

src/client/src/containers/SelectBackendFramework/index.tsx
```diff
@@ -77,6 +77,7 @@ class SelectBackEndFramework extends React.Component<Props> {
       <div className={styles.container}>
         {options.length > 0 && (
           <SelectOption
+            isFrameworkSelection={true}
```
streetcoder123, line 80: Can you explain what is the purpose of this prop `isFrameworkSelection={true}`?

src/client/src/containers/SelectFrameworks/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+      });
+      vscode.postMessage({
+        module: EXTENSION_MODULES.DEPENDENCYCHECKER,
+        command: EXTENSION_COMMANDS.GET_DEPENDENCY_INFO,
+        payload: {
+          dependency: "python"
+        }
+      });
+    }
+  }
```
streetcoder123, line 42: This is looking good. Another idea to consider (maybe for future) is to actually fire this post message to the vscode extension on every click to a different framework card instead just loading all of them only once when the component loads up. This would also be beneficial if the user adds a dependency and then changes card. In that case, he would get the updated information.

src/extension/src/constants.ts
```diff
@@ -205,6 +205,12 @@ export const CONSTANTS = {
   START_PORT: 9502,
   VSCODE_COMMAND: {
     OPEN_FOLDER: "vscode.openFolder"
+  },
+  DEPENDENCY_CHECKER: {
+    NODE: 'node',
+    PYTHON: 'python',
+    PYTHON3: 'python3',
+    PYTHON_LAUNCHER: 'py -3'
```
streetcoder123, line 213: @jasmingoh Why do you need to test with Python Launcher if you are already checking with both `python` and `python3` commands. Are there any cases missed by these two?

src/client/src/containers/SelectBackendFramework/index.tsx
jsondoo: > New prop `isFrameworkSelection` passed down to SelectableCards because this component is actually shared between all the pages, I only want the dependency info showing on the frameworks page

streetcoder123: You mean the component `SelectBackendFramworks` spans across multiple pages in the wizard?

SahilTara: I think he means the select option component is used by all pages.

src/extension/src/constants.ts
jasmingoh: https://github.com/microsoft/WebTemplateStudio/pull/758#issuecomment-503737686

tldr; `python` and `python3` commands are not consistent when user has both python 2.x and python 3.x installed.

src/client/src/containers/SelectBackendFramework/index.tsx
jsondoo: Yeah, `SelectableCard` is shared among Pages and Framework selection, I needed the cards to somehow know it was framework selection

streetcoder123: I see, that makes sense. Thanks for explaining that :)

src/client/src/App.tsx
jsondoo: done

src/client/src/containers/SelectFrameworks/index.tsx
jsondoo: yea, good idea

src/extension/src/utils/dependencyChecker.ts
jsondoo: done

src/extension/src/utils/dependencyChecker.ts
jsondoo: dine

src/extension/src/utils/dependencyChecker.ts
jsondoo: We think that this is okay as is but we could change it later

jsondoo:
Update 2:
More type safety changes per @SahilTara 
If-else changes per @piotrmark 
It looks a looot better now, so thanks everyone for the feedback

src/client/src/App.tsx
SahilTara: Hey @jsondoo think you forgot this one

streetcoder123 (dismissed):

src/client/src/App.tsx
jsondoo: oh man, whoops, thanks @SahilTara

SahilTara (approved):

streetcoder123 (approved):
Awesome :)

annielovesjs (approved):
Changes look good! Works well

Tanya0609:
It works properly. Great Work 👍 

I guess one thing I found out during my testing was: When I has no Node installed, it shows the "No Node detected message", then I installed Node. But at the same session, it is not able to redetect that Node has been installed. Only until I close the session and restart my extension. It detects the Node has been installed.

jasmingoh:
Tested on Mac and Linux, and it works!

jasmingoh:
@Tanya0609 the extension side will only be trigerred when the framework page loads again. Could you try that?

streetcoder123:
@jasmingoh The problem is not the extension but actually the VSCode terminal. When you install something say 'Node', the VSCode terminal does not detect that until VSCode is restarted. That is why, in order for it to detect an installation, complete VSCode needs to be shut down and started again.
Maybe we could ask the VSCode team about this.

Tanya0609 (approved):

piotrmark:
> @jasmingoh The problem is not the extension but actually the VSCode terminal. When you install something say 'Node', the VSCode terminal does not detect that until VSCode is restarted. That is why, in order for it to detect an installation, complete VSCode needs to be shut down and started again.
> Maybe we could ask the VSCode team about this.

That's very true, I'd say it's even deeper - when you install something, let's say Node, you have to restart PowerShell to pick it up.

piotrmark (approved):
Thanks again for your cooperation with our feedback! My pleasure to approve :D

PR #761, opened 2019-06-19
add full stack description

annielovesjs:
- Change: state that it is a full stack application on the welcome page
- test by running the application and seeing if the text on the front page is "Enter a project name and output path, then click next to get started on your full stack application."


jasmingoh (approved):
Change is showing up on the extension (highlighted), and the code looks good to me !

image

jsondoo (approved):
Looks good, maybe we could update the README as well? 👍

jcoc611-microsoft (approved):

src/client/src/containers/NewProject/index.tsx
```diff
@@ -54,7 +54,7 @@ const NewProject = ({
       <div className={styles.body}>
         <FormattedMessage
           id="newProject.body"
-          defaultMessage="Enter a project name and output path, then click next to get started."
+          defaultMessage="Enter a project name and output path, then click next to get started on your full stack application."
```
jcoc611-microsoft, line 57: Out of curiosity, why do we need a default message?

SahilTara: Hey, its actually used in localization to set the fallback value of the id (newProject.body). So if the value isn't defined for newProject.body in the json for french (example), it will give that message. It automatically populates the en.json with that value after running a command too.

PR #763, opened 2019-06-19
Cosmos Mongo Integration with Flask Templates

### LINKED ISSUE #762: Create Flask Templates to work with Cosmos Mongo
We need to integrate Flask templates with CosmosDB mongo database.

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/28595

**Changes**
- Add support for cosmos deployment using mongo database when Flask is used as a backend framework

**How to Check**
- Generate a project with Flask as the backend framework and any frontend framework and use cosmos deployment with mongoDB.
- If not already installed, install the following python packages:
`pip install flask`, `pip install pymongo`, `pip install python-dotenv`

- Try to run the generated project once the deployment is finished.
- The project should work appropriately.

closes #762


templates/Web/_composition/Flask/Feature.Flask.Azure.Cosmos.Mongo/server/mongo/settings.py
```diff
@@ -0,0 +1,14 @@
+import os
+from os.path import join, dirname, realpath
+from dotenv import load_dotenv
+
+#create .env file path
+dotenv_path = join(dirname(dirname(dirname(realpath(__file__)))), '.env')
+
+#load file from path
+load_dotenv(dotenv_path)
```
jcoc611-microsoft, line 9: Out of the scope of this PR, but it seems to me that using `.env` files for secrets is not the best security practice. Of course, it is difficult to not use this but also make the resulting app work in different environments. If the app is to be deployed in Azure we could use a Key Vault. At the very least we should add `.env` to a generated `.gitignore`

templates/Web/_composition/Flask/Page.Flask.List.AddRoutes/.template.config/template.json
```diff
@@ -10,7 +10,7 @@
     "wts.type": "composition",
     "wts.platform": "Web",
     "wts.version": "1.0.0",
-    "wts.compositionFilter": "$backendframework == Flask & groupIdentity == wts.Page.List"
+    "wts.compositionFilter": "$backendframework == Flask & groupIdentity == wts.Page.List & $feature != wts.Feature.Azure.Cosmos.Mongo & $feature != wts.Feature.Azure.Cosmos.SQL"
```
jcoc611-microsoft, line 13: A thought for later: how will this logic scale if we start adding more databases?

templates/Web/_composition/Flask/Feature.Flask.Azure.Cosmos.Mongo/server/mongo/settings.py
streetcoder123: I agree. To add .env to a .gitignore file, would it make sense to generate a .gitignore if we are not making a git repository for the generated project? Just wondering if it confusing to the user or is a normal behavior?

templates/Web/_composition/Flask/Page.Flask.List.AddRoutes/.template.config/template.json
streetcoder123: Good point. It would be a lot of clutter if we are going to use this logic. I am not sure but I think maybe if there is something like '$feature=None` if it even exists

jsondoo (approved):
can't comment much on flask code but deployment works, and cosmosDB is updated with list items 👍

Tanya0609 (approved):
not a flask experienced user, but it hook with database well, adding and deleting works properly

jcoc611-microsoft (approved):
we can iterate on the Flask style as suggestions come in from experts

templates/Web/_composition/Flask/Feature.Flask.Azure.Cosmos.Mongo/server/mongo/settings.py
piotrmark: Yeah, we could add TODO / readme item to explicitly say "hey, user, take care of secrets management" ;)

PR #775, opened 2019-06-22
Template test part 1: generate project script

annielovesjs:
- test by running yarn generate when you're in the extension directory and it should generate all the different permutations of the projects and output it into the template_test directory
- run yarn install and yarn start on one or two projects to make sure all the pages show up (should be one of each page for all project types)

(second part of test that will be in PR on monday will run yarn install, yarn start and yarn build for all the project types and throw errors if they exist and the files will be removed the test finishes)


streetcoder123:
Would you like to change the title of the PR. It has been set to the name of your branch by default :p

src/extension/src/scripts/generate-test.ts
```diff
@@ -0,0 +1,129 @@
...
+          .catch((error: Error) => {
+            throw Error(error.toString());
+          });
+      })
+      .catch(() => {
+        syncAttemptNum++;
+        return delay(3000).then(() => attemptSync(instance, syncAttemptNum));
+      });
+  }
+  if (syncAttemptNum >= CONSTANTS.API.MAX_SYNC_REQUEST_ATTEMPTS) {
```
piotrmark, line 62: NIT: I believe this will be always true, so this condition is redundant, but in general I think it would be nice to move this to the top to reduce nesting:
```suggestion
  if (syncAttemptNum >= CONSTANTS.API.MAX_SYNC_REQUEST_ATTEMPTS) {
    CoreTemplateStudio.DestroyInstance();
    throw new Error("too many failed sync requests");
  }
  return instanceObj...
  //Everything else
```

src/extension/src/scripts/generate-test.ts
```diff
@@ -0,0 +1,129 @@
...
+              (obj: { tags: { type: string }; name: string }) => {
+                if (obj.tags.type == "frontend") {
+                  frontend.push(obj.name);
+                } else if (obj.tags.type == "backend") {
+                  backend.push(obj.name);
+                }
+              }
+            );
+
+            function getPagesObj(
```
piotrmark: I think it would be more clear to move this function outside of attemptSync

src/extension/src/scripts/generate-test.ts
```diff
@@ -0,0 +1,129 @@
...
+                    });
+                  });
+                })
+                .catch(err => {
+                  console.log(err);
+                });
+              return pagesObj;
+            }
+
+            function generateProj(backend: string, frontend: string) {
```
piotrmark: Same here, can this function be declared outside?

src/extension/src/scripts/generate-test.ts
```diff
@@ -0,0 +1,129 @@
+import { CoreTemplateStudio } from "../coreTemplateStudio";
+import { CONSTANTS } from "../CONSTANTS";
+
+let instance: CoreTemplateStudio;
+let backend: string[] = [];
+let frontend: string[] = [];
```
piotrmark: Can we change names to plurals (backends, frontends) as these are arrays? :)

src/extension/src/scripts/generate-test.ts
```diff
@@ -0,0 +1,129 @@
...
+                  frontend.push(obj.name);
+                } else if (obj.tags.type == "backend") {
+                  backend.push(obj.name);
+                }
+              }
+            );
+
+            function getPagesObj(
+              instance: CoreTemplateStudio,
+              projType: string,
```
piotrmark: These parameters (instance, projType) will hide those declared above. Is this intentional?

src/extension/src/scripts/generate-test.ts
```diff
@@ -0,0 +1,129 @@
...
+                    pagesObj.push({
+                      name: page.name,
+                      identity: page.templateId
+                    });
+                  });
+                })
+                .catch(err => {
+                  console.log(err);
+                });
+              return pagesObj;
```
jcoc611-microsoft: I think this logic should work, but you're mixing sync and async code here. When you return `pagesObj` here it is undefined as the promise above has not been resolved. You can return the above promise instead and then wait for it in `generateProj`. You can verify this is an issue by looking at one of the generated projects; it has no pages.

src/extension/src/signalr-api-module/generateCommand.ts
```diff
@@ -1,5 +1,5 @@
 import { CoreTemplateStudioApiCommand } from "./coreTemplateStudioApiCommand";
-import { CONSTANTS } from "../constants";
+import { CONSTANTS } from "../CONSTANTS";
```
jsondoo: Just wondering, will this break Linux again? @jasmingoh

src/extension/src/signalr-api-module/syncCommand.ts
```diff
@@ -15,7 +14,6 @@ export class SyncCommand extends CoreTemplateStudioApiCommand {
         this.commandPayload.payload!.path
       )
       .catch((error: Error) => {
-        Logger.appendLog("EXTENSION", "error", error.message);
```
jsondoo, line 18: why did we remove this log?

src/extension/src/coreTemplateStudio.ts
```diff
@@ -46,24 +46,23 @@ export class CoreTemplateStudio {
...
     if (platform === CONSTANTS.API.WINDOWS_PLATFORM_VERSION) {
       executableName += ".exe";
     }
 
-    let apiPath = vscode.Uri.file(
-      path.join(extensionPath, "src", "api", platform, executableName)
-    ).fsPath;
-
-    let apiWorkingDirectory = path.join(
+    let apiPath = path.join(
```
streetcoder123, line 56: What is `vscode.Uri.file`? Has it been removed intentionally?

src/extension/src/scripts/generate-test.ts
```diff
@@ -0,0 +1,126 @@
...
+            frontends.forEach(frontendFrameWork => {
+              backends.forEach(backendFramework => {
+                prevPromise = prevPromise.then(() =>
+                  generateProj(backendFramework, frontendFrameWork)
+                );
+              });
+            });
+
+            prevPromise.then(() => {
+              console.log("done");
```
streetcoder123: Instead of generating each permutation of framework one after the other, I think it could be faster and more readable to use a `Promise.all` instead. In that case, `prevPromise` would be an array of promises with different permutations of frontend and backend frameworks. Then we can use `Promise.all(prevPromise)` to resolve all the promises and say `done`.
Does that make sense or I am thinking something wrong?

src/extension/src/scripts/generate-test.ts
streetcoder123: @annielovesjs this is a pretty good suggestion. Have this condition at top to fail. Then you would know that number of sync attempts is less than max allowed for sure and you won't have to wrap that in the `if` clause.

annielovesjs: actually, this condition is only true when we fail to connect to core and continuously try to sync until we reach a max amount of attempts which generally doesn't happen if we connect to core properly but I did make new changes to wrap that in an else clause instead of having a separate if statement

annielovesjs: @piotrmark @njindal239 true, good catch! It'll definitely make it a lot cleaner. I'll make the changes now

src/extension/src/scripts/generate-test.ts
annielovesjs: @piotrmark thanks for the suggestion! :) just changed it and it looks a lot cleaner now

src/extension/src/coreTemplateStudio.ts
annielovesjs: I actually removed it for testing purposes since I couldnt use the vs code module because the test runs without running the extension. I forgot to add it back - thanks for reminding me! I just added it back

src/extension/src/coreTemplateStudio.ts
```diff
@@ -42,27 +42,27 @@ export class CoreTemplateStudio {
...
     }
 
     if (platform === CONSTANTS.API.WINDOWS_PLATFORM_VERSION) {
       executableName += ".exe";
     }
 
-    let apiPath = vscode.Uri.file(
-      path.join(extensionPath, "src", "api", platform, executableName)
-    ).fsPath;
+    // if (context) {
```
SahilTara: Should remove this commented out code.

src/extension/src/coreTemplateStudio.ts
```diff
@@ -42,27 +42,27 @@ export class CoreTemplateStudio {
     let platform = process.platform;
     let executableName = CONSTANTS.API.BASE_APPLICATION_NAME;
     let extensionPath;
+    let apiPath;
```
SahilTara: why are we defining apiPath up here when it can be defined at line 63?

```ts
let apiPath = path.join(extensionPath, "src", "api", platform, executableName);
```

src/extension/src/scripts/generate-test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+const delay = (time: number) => {
+  return new Promise(function(resolve) {
+    setTimeout(resolve, time);
+  });
+};
+
+let getPagesObj = (
+  instance: CoreTemplateStudio,
+  frontend: any,
```
SahilTara: nit: frontend: string, backend: string

src/extension/src/scripts/generate-test.ts
```diff
@@ -0,0 +1,126 @@
...
+                frontends.push(obj.name);
+              } else if (obj.tags.type == "backend") {
+                backends.push(obj.name);
+              }
+            }
+          );
+
+          frontends.forEach(frontendFrameWork => {
+            backends.forEach(backendFramework => {
+              prevPromise = prevPromise.then(() =>
```
SahilTara: what is the goal of this prevPromise variable?  Mainly asking since on line 103 there is a console.log("done");

src/extension/src/scripts/generate-test.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  return instanceObj
+    .sync({
+      port: instance.getPort(),
+      payload: { path: CONSTANTS.API.DEVELOPMENT_PATH_TO_TEMPLATES },
+      liveMessageHandler: value => {
+        value;
+      }
+    })
+    .then(() => {
```
SahilTara: Edit: added an async await. 
i feel like this nesting is unnecessary, we could do something like the following:
```ts
return instanceObj
.sync({
  // do stuff
})
.then(async () => {
  return await instance.getFrameworks(projType);
})
.then(frameworks => {
  //do stuff here.
})
.catch(// handle errors, could be syncattempt or something else, make sure to catch it properly.)
```
Feel like this is a bit more readable? What do you think?

SahilTara (commented):
Left a few questions/suggestions. Overall good job.

src/extension/src/scripts/generate-test.ts
jcoc611: @njindal239 unfortunately, it is necessary that one generation of the framework happen after the other. This is because CoreTS gets corrupted if you try generating one project before another generation is done.

src/extension/src/coreTemplateStudio.ts
jcoc611: This is my bad. I'm used to a different project's style guidelines that requires all variables be declared at the beginning of code blocks. For this project feel free to change it if it is more readable that way.

src/extension/src/signalr-api-module/syncCommand.ts
annielovesjs: I removed it because it originally caused my test to error out since it uses the module vscode but my test runs independently from the extension and it was also inconsistent with generate command

I moved the logger to the catch block of attemptsync in launchexperience so it should achieve the same purpose as including it in syncCommand

src/extension/src/scripts/generate-test.ts
SahilTara: Oh wait is this essentially generating an execution chain of generations? So it would be like generate#1, after that generate#2, etc.. until we reach the eventual console.log of done?

jcoc611-microsoft: Yeah, the last console.log is needed so the last promise in the chain is resolved

src/extension/src/coreTemplateStudio.ts
jcoc611-microsoft: This one is my bad, used to a different style requirement of declaring all variables at the start of code blocks for readability. Feel free to change here.

src/extension/src/scripts/generate-test.ts
jcoc611-microsoft: @njindal239  Unfortunately, we must generate projects one after the other. This is because CoreTS gets corrupted if we try to generate a project before a previous generation is done. Thus, we must wait for one project to be generated before we continue to generate another. We achieve this by having a promise chain.

`Promise.all` would try to generate all projects at the same time (would send GENERATE commands one after the other before receiving replies from CoreTS through signalr)

src/extension/src/coreTemplateStudio.ts
jcoc611-microsoft: `path.join` takes care of the path for us. We need to remove dependencies on `vscode` here so we can run the generate script by itself, outside the extension.

src/extension/src/scripts/generate-test.ts
annielovesjs: That's a good idea! It definitely helps make the code more readable. I'll add it in right now :) Thanks!

src/extension/src/scripts/generate-test.ts
streetcoder123: I agree. I did not know that we need to have one project complete its generation before we can start the second one. But that's a little weird to me why CoreTS behaves that way.

src/extension/src/coreTemplateStudio.ts
annielovesjs: just changed it so its all declared in one line :)

annielovesjs:
@njindal239 @SahilTara @jsondoo @piotrmark @jcoc611-microsoft Thank you so much for all your feedback and suggestions! I've incorporated all the changes and it now looks a lot better than before :) If you guys test it now, everything should generate properly and if you run webts and try to generate a project it shouldn't interfere with anything either!

.vscode/launch.json
```diff
@@ -0,0 +1,17 @@
...
+      "type": "node",
+      "request": "launch",
+      "name": "Launch Program",
+      "program": "${workspaceFolder}\\src\\extension\\src\\coreTemplateStudio.ts",
+      "outFiles": [
+        "${workspaceFolder}/**/*.js"
+      ]
+    }
+  ]
+}
```
streetcoder123: Does anyone know what is this file for `.vscode/launch.json`?

src/extension/src/scripts/generate-test.ts
```diff
@@ -0,0 +1,119 @@
...
+let getPagesObj = (
+  instance: CoreTemplateStudio,
+  frontend: string,
+  backend: string
+) => {
+  return instance
+    .getPages(projType, frontend, backend)
+    .then(pages => {
+      pagesObj = [];
+      pages.forEach((page: { name: any; templateId: any }) => {
```
SahilTara: we can simplify this even more using:
```ts
return pages.map((page: { name: string; templateId: string }) => {
  return {name: page.name, identity: page.templateId};
}
```

Eliminating the need for pagesObj.

.vscode/launch.json
SahilTara: Pretty sure this is like a f5 experience within the root directory. Maybe for local testing or something?

annielovesjs: it was generated when I was trying to debug the application but overall, it doesn't affect the webts generation experience or my script so I removed it in the new commit

src/extension/src/scripts/generate-test.ts
annielovesjs: good call! I just made the changes

src/extension/src/scripts/generate-test.ts
SahilTara: The reason it works that way is because it (the api) was meant to be a 1 connection at a time thing, so it was developed limiting it to 1 connection. This can be changed fairly easily though if it proves to be a huge hinderance.

SahilTara (dismissed):

src/extension/src/scripts/generate-test.ts
```diff
@@ -0,0 +1,113 @@
...
+      frontends.forEach(frontendFrameWork => {
+        backends.forEach(backendFramework => {
+          prevPromise = prevPromise.then(() =>
+            generateProj(instance, backendFramework, frontendFrameWork)
+          );
+        });
+      });
+
+      prevPromise.then(() => {
+        console.log("project generation complete");
```
jcoc611-microsoft, line 95: Could you add `CoreTemplateStudio.DestroyInstance();` here so the process cleans up after generation? Otherwise it will keep executing even after generation is done.

annielovesjs: Yeah true, I forgot to destroy the instance. Thanks for the reminder! Just committed the change

jcoc611-microsoft (approved):
Great work!

SahilTara (approved):

jsondoo (approved):
good work, annie :)

piotrmark (approved):
Beautiful, you nailed it :) Congrats!

PR #782, opened 2019-06-24
Add requirements.txt file for installing pip Dependencies

### LINKED ISSUE #776: Flask Feedback - Generate a requirements.txt file to install the required python Dependencies
Flask Feedback by @jcoc611-microsoft 

Instead of instructing the user to manually install the python dependencies, plop all the required dependencies into a requirements.txt file that could be run using the `pip` command directly to install all the required dependencies.

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/29397

**Changes**
- This PR adds a requirements.txt file in a generated Flask application such that the user could just run the file to install all the required dependencies instead of having to install all the python dependencies manually.

**How to Check**
- Generate a Flask project with Cosmos DB selected
- In the generated project, there should be a `requirements.txt` file that has all the required `pip` dependencies
- Run `yarn install-requirements` should run that file to install all the dependencies
- Running `yarn start` should start the project appropriately.

closes #776


jsondoo (dismissed):

piotrmark (dismissed):

annielovesjs (dismissed):

Tanya0609 (approved):
It works. But also want to point out: it also introduces error message when do "npm start", since I have python3 installed but no python installed
python

jcoc611-microsoft:
> It works. But also want to point out: it also introduces error message when do "npm start", since I have python3 installed but no python installed

This will be problematic later when we try to integrate the tests in the pipeline. The easiest way to test is to verify there is no output in stderr. We should think a bit more how to fix this.

streetcoder123:
@Tanya0609 @jcoc611-microsoft Yeahh, this is a problem introduced when we try to introduce OS compatibility such that both commands run. So, for windows, since `python3` is not a recognized command on windows, it would first give an error before running the next command that works.

Currently, I am not able to think of a better way but I am thinking about it. Please share if you have some ideas to go about it.

jsondoo (approved):

piotrmark (approved):

PR #787, opened 2019-06-24
OS Compatibility for running Flask Application

### LINKED ISSUE #786: Running Flask server should be compatible with all OSes
Different OSes have different commands for running `python` scripts. For instance, windows command for running python script is `python script.py` whereas for macOS, it is `python3 script.py` for version 3 of python.
We need our server to boot up on all OSes.

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/29403

**Changes**
- Allow both `python` and `python3` to boot up the server so that all Oses can start the Flask server

closes #786


piotrmark (approved):

jasmingoh (approved):

jcoc611-microsoft (approved):
It looks okay. I think in most cases it will work as expected. Perhaps if both `python3` and `python` are installed, and the script fails with an exception, then it will be called twice. Seems like an acceptable edge case though

PR #792, opened 2019-06-25
Flask - Function and Variable names should reflect the PEP8 Python Standards

### LINKED ISSUE #785: Change all the function and variable names to reflect the PEP8 Python Standards
All the function and variable names used within the Flask templates should respect the PEP8 coding style for Python projects.
https://www.python.org/dev/peps/pep-0008/#function-and-variable-names

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/29399

**Changes**
- Adapt all the functions and variables in Flask templates to respect PEP8 standards

**Note**: Please look for any places that I might have missed to incorporate similar standards.

**How to Test**
- This PR touches many Flask templates. So, the best way to test this is to generate all 3 types of projects (with Flask as the backend framework):
    - All pages without any CosmosDB
    - All pages with CosmosDB Mongo
    - All pages with CosmosDB SQL
- Try to run the project and it should work appropriately.

closes #785


templates/Web/_composition/Flask/Page.Flask.List.AddRoutes.WithCosmos.SQL/server/server_postaction.py
```diff
@@ -12,37 +12,37 @@
...
 
 //{[{
-sqlDatabaseObj = SQLObj()
+sql_database_obj = SQLObj()
 
 # List Endpoints
 @app.route(CONSTANTS['ENDPOINT']['LIST'])
-def getList():
+def get_list():
     queryStr = {
```
jasmingoh: should queryStr be underscore separated too?

streetcoder123: Missed that one. Thanks for pointing it out :)

templates/Web/_composition/Flask/Page.Flask.MasterDetail.AddRoutes/server/server_postaction.py
```diff
@@ -10,9 +10,9 @@
 
 //{[{
 # MasterDetail Page Endpoint
-@app.route(CONSTANTS['ENDPOINT']['MASTERDETAIL'])
-def getMasterDetail():
+@app.route(CONSTANTS['ENDPOINT']['MASTER_DETAIL'])
+def get_msater_detail():
```
jsondoo: typo here oops

jsondoo:
@njindal239 will there be more pep8 stuff after this?

streetcoder123:
@jsondoo I am planning to do some module name changes. I did not do that in this PR because it would have led to huge diffs since it would have meant renaming file with changes.

SahilTara (approved):

templates/Web/_composition/Flask/Page.Flask.AddSampleData.ForAllExceptList/server/sampleData$wts.Page_gpostaction.py
```diff
@@ -2,19 +2,19 @@
 import datetime
```
SahilTara, line 2: Just wondering if the datetime causes problems. Since theres no differentiator between this and the shortLoremIpsum. When you generate grid and list for example only 1 short lorem ipsum is generated right?

streetcoder123: That's a good catch. It does combine the macros. But currently, I am not able to think of a way to separate that. I mean we need to have some kind of default text between macros to avoid combining.
Hmm, need to think about it

streetcoder123: Actually, I can fix it. We need to swap the `short_lorem_ipsum` and `long_lorem_ipsum` since `long_lorem_ipsum` is just used for grid/masterDetail sample data. This is not a generalized good solution but in consideration to the constraints in CoreTS, this is the best we can do.
I will make a small PR for that

jasmingoh (approved):

annielovesjs (commented):
looks good!

jasmingoh:
While on this branch, we found an issue on Mac, please refer here #796

annielovesjs (approved):

SahilTara:
@jasmingoh https://github.com/microsoft/WebTemplateStudio/issues/796#issuecomment-505683633 I left a comment on the issue, but I feel like it has to do with pip install.

jsondoo (approved):

piotrmark (approved):
Yep, looks more like Python now :D

PR #793, opened 2019-06-25
Dependency Checker clean up

jsondoo:
Small refactoring on `DependencyInfo` component
1) `DependencyInfo` takes in just frameworkName
2) `DependencyInfo` uses frameworkName to get the dependency info and links (privacy link will be used later), and looks up the store to get the installation state
3) Renaming some types

**How to test**
Behaviour for the dependency checker should remain the same


src/client/src/containers/DependencyInfo/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+interface IDependency {
+  [key: string]: string;
+}
+
+interface IDependencies {
+  [key: string]: IDependency;
+}
+
+const dependencies: IDependencies = {
+  NodeJS: {
```
jcoc611-microsoft, line 46: This is great! Do you think it would be possible to move this metadata to `backendframeworks.json` on the `templates/_catalog` folder? Not super important but it would be easier to keep track of for future frameworks. I know there isn't a "Python" backend but maybe it could be possible with tags? If it's too much work this is fine too.

jsondoo: Initially I was considering putting dependencies as a new attribute for each framework in the catalog but seemed repetitive.

I will look into tags

src/client/src/containers/DependencyInfo/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+    let { frameworkName, intl, dependenciesStore } = this.props;
+    let dependency: IDependency | undefined = frameworkNameToDependencyMap.get(
+      frameworkName
+    );
+
+    if (dependency === undefined) {
+      return null; // don't render anything
+    }
+
+    const dependencyName: string = dependency.displayName;
```
SahilTara: think you could do something like:
```ts
const { downloadLink, privacyStatementLink, displayName: dependencyName} = dependency;
```

Why not call the displayName the dependencyName and call the dependencyName the commandName? or something. Would allow you to destructure cleaner, and also destructure "dependency.commandName" for the  installed variable.

src/client/src/containers/DependencyInfo/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+    defaultMessage: " not detected. Click to install."
+  },
+  iconAltMessage: {
+    id: "dependencyChecker.iconAltMessage",
+    defaultMessage: "Icon for dependency checker"
+  }
+});
+
+interface IDependency {
+  [key: string]: string;
```
SahilTara: Feel like you should make this a bit more solid. Like dependencyName, displayName, downloadLink, privacyStatementLink.

src/client/src/containers/DependencyInfo/index.tsx
jsondoo: agreed, that sounds better

src/client/src/containers/DependencyInfo/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+    let { frameworkName, intl, dependenciesStore } = this.props;
+    let dependency: IDependency | undefined = frameworkNameToDependencyMap.get(
+      frameworkName
+    );
+
+    if (dependency === undefined) {
+      return null; // don't render anything
+    }
+
+    const { dependencyName, downloadLink, privacyStatementLink } = dependency;
```
SahilTara: Just wondering where is the privacyStatementLink (Didn't notice earlier) used? also you can add commandName here too, makes the code a bit cleaner.

jsondoo: It's not used atm, fixed!

src/client/src/containers/DependencyInfo/index.tsx
jsondoo: I don't think I will be moving it to `backendframeworks.json` in this PR, but i agree it would be easier to update in the future

Do you think maybe it would make sense to have a separate `dependency.json` in the catalog? @jcoc611-microsoft

jcoc611-microsoft: I think for this to work you would have to also change CoreTS to pick it up? Not sure, but I'm alright with the way you have it for now.

src/client/src/containers/SelectableCard/index.tsx
```diff
@@ -95,20 +93,9 @@ const SelectableCard = ({
...
-              installed={dependencies.python.installed}
-            />
-          ) : (
-            <DependencyInfo
-              frameworkName={"NodeJS"}
-              installed={dependencies.node.installed}
-            />
-          ))}
+        {isPreview && isFrameworkSelection && selected && (
+          <DependencyInfo frameworkName={option.internalName} />
```
streetcoder123, line 97: Just for curiosity, I don't know where do we get the `option` object from?

src/client/src/mockData/mockVsCodeApi.ts
```diff
@@ -40,7 +40,7 @@ const mockVsCodeApi = () => ({
               command: EXTENSION_COMMANDS.GET_DEPENDENCY_INFO,
               payload: {
                 dependency: "python",
-                installed: false
+                installed: true
```
streetcoder123, line 43: Any specific reason to change the mock data?

streetcoder123 (commented):
It looks great. Just a couple of questions out of curiosity.

src/client/src/mockData/mockVsCodeApi.ts
jsondoo: No specific reason, but on my computer I usually have node installed and no python so I can check that through the extension. On the browser, I can check the opposite (node not installed, and python installed warnings)

src/client/src/containers/SelectableCard/index.tsx
jsondoo: It's grabbed from the store in `SelectBackendFramework` and `SelectFrontendFramework` and then passed down to here, it contains information about the frameworks like author, longdescription, version, licenses, etc

streetcoder123 (dismissed):
Code looks good! I have tested this on Windows and it works. It would be great if someone could also test it on other OSes.
Great work :)

SahilTara (dismissed):

src/client/src/containers/DependencyInfo/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  privacyStatementLink: string;
+}
+
+interface IDependencies {
+  [key: string]: IDependency;
+}
+
+const dependencies: IDependencies = {
+  NodeJS: {
+    commandName: "node", // command used on the extension side
```
jasmingoh: The extension side is still not using this commandName, right?

src/client/src/containers/DependencyInfo/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+    dependencyName: "Node",
+    downloadLink: "https://nodejs.org/en/download/",
+    privacyStatementLink: "https://nodejs.org/en/about/privacy/"
+  },
+  Python: {
+    commandName: "python",
```
jasmingoh: Related to the previous comment, we may need to change more stuff (not sure where yet) given that the extension side checks for python using different commands (i.e. `python`, `python3` and `py -3`)

jsondoo: Yea I agree, I really didn't want to touch the extension stuff as well in this pr

Basically `commandName` is same as the attribute for the payload.dependency that we pass back and forth between client and extension, on the extension side you take this `commandName` (payload.dependency), check it, then run it directly.

SahilTara: Pretty sure command name is just a dictionary key. It isnt actually executed. We group all of the python commands as python and all node as node. It doesn't seem to be sending any requests to the extension. I assume this means the data already exists. Maybe a better name would help you? dependencyStoreKey or something?

SahilTara: Unless this code is meant to send requests but doesn't yet in which case you'd be right.. don't see any dispatching here though. Think this is just meant to grab the data we want and map it to the right framework using the dependencyKey (commandName)

jsondoo: I needed a separate attribute `displayName` that's capitalized to display it

src/client/src/containers/DependencyInfo/index.tsx
jsondoo: This object is separate from the extension side. 
The thing is that right now I use the payload I get back from the extension directly in the store, so it's lowercase. I think we can discuss how to change it but probably need some mapping from what I send you from the client to the command you actually run on the extension side

src/client/src/containers/DependencyInfo/index.tsx
jsondoo: sorry for the confusion, this `commandName` isn't the one actually executed, but it is same as the one you execute from the payload I send from the client

src/client/src/containers/DependencyInfo/index.tsx
jasmingoh: Ok, hm, what about changing the comment to "message received from the extension side"? because the commandName variable name sounds like it is something that will be executed

jsondoo: Yea, naming is hard

jasmingoh: Talked to @jsondoo about this: 
I will create a map (or just create extra constants) on the extension side for the dependencyName and actual command that will be run. The extension will be checking dependencyName from the client, and also returning dependencyName ("Python" or "Node") to the client

jasmingoh (approved):
Looks good!

streetcoder123 (approved):

SahilTara (approved):

PR #794, opened 2019-06-25
Global .gitignore file which can then be used by both frontend and backend frameworks through postactions

### LINKED ISSUE #769: Backend Frameworks should also add to generated .gitignore File
Currently, when we generate a project, only the frontend framework adds '.gitignore' file to the project. However, we should add more content to '.gitignore' through the backend frameworks as well.
For instance, '.env' file generated should also be part of .gitignore file

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/29402

**Changes**
- This PR creates a global .gitignore file for any Full Stack Application. The frontend and backend frameworks can then add to the file using postactions.
- Also addresses the gitignore postaction not working for Node.
- Creates a new gitignore postaction for Angular since it was not already present.

**How to Test**
- This PR touches all the frontend and backend frameworks.
- Generating projects with all frameworks would be a fairly robust check.

closes #769


jcoc611-microsoft (dismissed):

SahilTara (dismissed):

jsondoo:
> * Generating projects with all frameworks would be a fairly robust check.

@annielovesjs only if we had your template test part 2 here haha

templates/Web/Projects/VueDefault/_postaction.gitignore
```diff
@@ -1,12 +1,10 @@
...
 
-# Hide API Keys
-.env
+//{[{
+.DS_Store
+
+/build
 
 # production
 /server/build
```
jsondoo: Should this be here? seems like this is duplicated in .gitignore when I generate a project with a backend

streetcoder123: Good catch! This should not be here.

jsondoo (approved):

Tanya0609 (approved):

jasmingoh:
Not sure if it is intended, but the structure of the generated gitignore file seems to be a little different for each frontend framework (React, Angular and Vue)

E.g. In Angular, `#production` is before `#testing`, but in React `#production` is before `#dependencies`
For Vue, `.DS_Store` is not grouped in #misc like how it is in React and Angular

jcoc611-microsoft (approved):

PR #803, opened 2019-06-26
Abstract Database Queries to CosmosDB to individual files

### LINKED ISSUE #802: Flask Cosmos - Abstract all Database Queries to individual files
As it is also done by the Node application, Flask application should also abstract away all the database query part into separate files.

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/29404

**Changes**
- The Routes for Database queries to cosmosDB have been abstracted to their individual files for both mongo and sql.

**How to Check**
- Generate a Flask project with all pages selected once with MongoDB and then with SQL.
- Both the projects should work appropriately.

closes #802


templates/Web/_composition/Flask/Feature.Flask.Azure.Cosmos.Mongo/server/mongo/mongo_service.py
```diff
@@ -0,0 +1,40 @@
...
+    data = request.get_json()
+    list_item = {'text': data['text']}
+    created_item = list_items.insert_one(list_item)
+    return make_response(
+        jsonify(
+            {'_id': str(created_item.inserted_id), 'text': list_item['text']}
+        ), 201
+    )
+
+def destroy(id):
```
SahilTara, line 24: why not just do 
```py
result = list_items.delete_one(query_str)

if result.deleted_count == 0:
  # return error
else: 
  # return comment deleted
```

jasmingoh:
Our method of validation CosmosDB name is currently not working. Please comment out this chunk of code when testing for this PR. https://github.com/microsoft/WebTemplateStudio/blob/dev/src/extension/src/azure/azure-cosmosDB/cosmosDbModule.ts#L342-L347

Make sure to put in a unique name when you're adding CosmosDB!

templates/Web/_composition/Flask/Feature.Flask.Azure.Cosmos.Mongo/server/mongo/mongo_service.py
```diff
@@ -0,0 +1,40 @@
+from flask import jsonify, make_response, request
+from .mongo_client import *
```
jasmingoh: Talked to @njindal239 about `import *` since it is usually discouraged. @njindal239 will address it in another PR.

jasmingoh (approved):

jasmingoh (approved):
Talked to @njindal239 about `import *` since it is usually discouraged. @njindal239 will address it in another PR.

SahilTara:
Wait why is it not working? Did something break or change in the client side/azure side? @jasmingoh

templates/Web/_composition/Flask/Feature.Flask.Azure.Cosmos.Mongo/server/mongo/mongo_service.py
streetcoder123: I shall be refactoring this as part of a feedback PR. Thanks for the suggestion.

streetcoder123:
@SahilTara When we fill out the form for Azure CosmosDB, it asks for the Account Name. The validation for Account Name calls an API on Azure which never returns. 
It seems to us there is some problem with the Azure side. We are still investigating it.

jsondoo:
Yea I think Azure SDK is broken

jasmingoh:
@SahilTara We have not change anything that validate the cosmosDB name, so we are guessing that it is Azure's problem. I've contacted Microsoft Support about it, still waiting for their response..

jcoc611-microsoft:
If we are going to update this in a refactor PR coming soon, does it make more sense to close this one and wait for the one with refactored code?

streetcoder123:
@jcoc611-microsoft We can do that. However, since the refactoring PR involves a lot of small changes relating to code style here and there, I think it would be good to incorporate all those small changes at once in a separate PR

jcoc611-microsoft (approved):
Seems good, waiting for future PR with refactor

SahilTara (approved):

piotrmark (approved):

PR #804, opened 2019-06-26
Enable resize of summary page right side bar

annielovesjs:
- made it so that once you change the width of styling on .rightViewCroppedSummaryPage, the sidebar width changes only for the summary page
- test it out by changing the width of .rightViewCroppedSummaryPage, build it, and then launch the extension to check out your changes on the summary page. Make sure only the width for the summary page has changed


SahilTara:
Just curious whats the end goal of this PR? Is it to make it easier to resize stuff once top navbar has replaced left navbar?

annielovesjs:
@SahilTara Yup :) The plan is to make the right sidebar a lot bigger on the summary page since we're also replacing all the content on the page itself.

src/client/src/containers/RightSidebar/index.tsx
```diff
@@ -147,7 +147,10 @@ class RightSidebar extends React.Component<Props, IRightSidebarState> {
         {pathname !== ROUTES.PAGE_DETAILS && (
           <div
             role="complementary"
-            className={classNames(styles.container, styles.rightViewCropped)}
+            className={classNames(styles.container, styles.rightViewCropped, {
+              [styles.rightViewCroppedSummaryPage]:
+                pathname === ROUTES.REVIEW_AND_GENERATE
```
jsondoo, line 154: Looks good!

NIT: if we want to avoid `!important` we could just toggle the styles
```
[styles.rightViewCropped]: pathname !== ROUTES.REVIEW_AND_GENERATE,
[styles.rightViewCroppedSummaryPage]: pathname === ROUTES.REVIEW_AND_GENERATE
```

piotrmark (changes requested):
+1 for Jason's comment - in general we should avoid using !important when possible - it is not the best practice

src/client/src/containers/RightSidebar/index.tsx
annielovesjs: Turns out I just needed to move the .rightViewCroppedSummaryPage below .rightViewCropped to be able to remove !important and have it work due to the css hierarchy. It should work now without important now!

annielovesjs:
@piotrmark I removed the need for important by moving .rightViewCropped above .rightViewCroppedSummaryPage  so .rightViewCroppedSummaryPage has higher priority

src/client/src/containers/RightSidebar/index.tsx
jsondoo: Okay, great! I think it might still be helpful to have the toggle so it's really clear which styles are being applied on the last page, rather than trying to figure it out from the order of css, but this is up to you

annielovesjs:
I am closing it until the PR for moving the nav bar to the top is in and then I'll reopen this and merge it

src/client/src/containers/RightSidebar/index.tsx
```diff
@@ -147,7 +147,10 @@ class RightSidebar extends React.Component<Props, IRightSidebarState> {
         {pathname !== ROUTES.PAGE_DETAILS && (
           <div
             role="complementary"
-            className={classNames(styles.container, styles.rightViewCropped)}
+            className={classNames(styles.container, styles.rightViewCropped, {
```
piotrmark, line 150: I'm a little bit confused here: we're conditionally adding rightViewCroppedSummaryPage class here in addition to rightViewCropped.
Now, rightViewCroppedSummaryPage has only `width: 400px`, but rightViewCropped has also `width: 400px`.
So, what value does it bring?
Maybe I missed something ;)

annielovesjs: oh! It' .rightViewCropperSummaryPage is useful for when we want to resize in the future - for ex. once the top nav bar is up, we can change the value of width to 600px or a bigger width

piotrmark: Ok, thanks :)

piotrmark (approved):

piotrmark (changes requested):
I think we should conditionally swap classes instead of relying on ordering, as @jsondoo suggested :)

annielovesjs:
@piotrmark I can do that but I also need the styling from rest of the styling from .rightViewCropped for the summary right side page. 

I can change it such that theres a separate class called .rightViewCroppedWidth where it only specifies the width for the right side bar in general and toggle it to .rightViewCroppedSummaryPage when pathname === ROUTES.REVIEW_AND_GENERATE

Is that the direction you would recommend I take?

piotrmark:
> @piotrmark I can do that but I also need the styling from rest of the styling from .rightViewCropped for the summary right side page.
> 
> I can change it such that theres a separate class called .rightViewCroppedWidth where it only specifies the width for the right side bar in general and toggle it to .rightViewCroppedSummaryPage when pathname === ROUTES.REVIEW_AND_GENERATE
> 
> Is that the direction you would recommend I take?

Ok, so I'd suggest to take the width out of rightViewCropped class and create new class, let's say rightViewCroppedNonSummaryPage (figure out better name :D) with just the width and then conditionally apply this one or rightViewCroppedSummaryPage

piotrmark (approved):
Thanks!

jsondoo (approved):
ship it

streetcoder123 (approved):

PR #807, opened 2019-06-27
Flask - short_lorem_ipsum is defined twice if both List and Grid pages are Selected

### LINKED ISSUE #795: Flask - short_lorem_ipsum is defined twice if both List and Grid pages are Selected
If both List and Grid pages are Selected, `short_lorem_ipsum` is defined twice in the sampleData file. This is due to the combining of macros.

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/29699

**Changes**
- Decouple the macros such that they do not get combined.

closes #795


jcoc611-microsoft (approved):
Although the fix works, this seems to me like it is a bit fragile. It relies on the order of statements for the post action logic of deleting duplicate lines, but this might not be obvious to people coming in to the project later. It would thus likely break after other changes. Maybe if all pages use this data we can have it on a shared Flask file instead?

jsondoo (approved):

annielovesjs (approved):

PR #808, opened 2019-06-27
summary page and modal works

Tanya0609:
Task:https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/WTS%20Team/Intern%20GitHub/Y19-VS/06

This change is to make current summary page (middle panel) as the expected screenshot. (The information originally in summary page is now showing on the right side bar, which is capture by annie's PR).

**How to test:**
1. generate any project type with any page
2. at summary page, clicks on **View Licenses** button, it opens a modal and contains all the corresponding licenses
3. at summary page, clicks on **Give Feedback** button, it should redirect you to GitHub Issue page

**New Summary Page from Designer:** ( the backgroud image in this design is from the internet)
summary page

**Expected:**
viewLicenses


streetcoder123:
@Tanya0609 Firstly, great work. Looking at the text inside the modal, would it make more sense to have a shorter and vertical modal instead of a horizontal one. I think that would make the text look a lot nicer. Thoughts?

SahilTara:
Yeah @njindal239 I agree. @Tanya0609 I also feel like there's a lot of unused space. Will there be something that goes there instead? Or would it be better to shrink the modal or change the layout?

jsondoo (commented):
Good work :)

Do you think we could also align the text with the button on the left?
image

Tanya0609:
@njindal239 @SahilTara Yes, we noticed that extra space there too. I consulted with Chrisina, it is okay so far. But we are going to do another PR to change the modal css.

Since modal component is shared among all the modals in the extension, in order to update the size for my modal, it will be need to decouple the css with other modals, which may lead to the redesign of the modals.

src/client/src/containers/ReviewAndGenerate/index.tsx
```diff
@@ -1,120 +1,82 @@
...
 
-const mapDispatchToProps = (
-  dispatch: ThunkDispatch<AppState, void, RootAction>
-): IDispatchProps => ({
-  openCosmosDbModal: () => {
-    dispatch(ModalActions.openCosmosDbModalAction());
-  },
-  openAzureFunctionsModal: () => {
-    dispatch(ModalActions.openAzureFunctionsModalAction());
+const mapDispatchToProps = (dispatch: any): IDispatchProps => ({
```
streetcoder123, line 73: Why did you change the type of dispatch to `any`?

Tanya0609:
closed accidently, thus reopen

jcoc611-microsoft:
This is gonna have merge issues with Jason's modal refactor I think. You both should sync so the two are compatible

src/client/src/containers/ReviewAndGenerate/index.tsx
Tanya0609: Good catch, I was just following another Modal's code and avoid unnecessary libraries, but it will be better to use "dispatch: Dispatch<ModalActionType>". Will update my PR.

Tanya0609:
I will close this PR so far that I can address Nipun's comment and add resize into this PR too. By the time I reopen this PR, Jason's PR should be merged and my PR will compatible with his change.

PR #809, opened 2019-06-28
Privacy Modal for Dependency Checker

jsondoo:
This changes the dependency checker to show a privacy modal when it's clicked since we direct our users to a non-Microsoft link such as python/node download link.

Changes:
- `PrivacyModal` component (should be similar to Tanya's licenses modal #808)
- Redux changes for the `openModal` store, so that we can pass data now when we dispatch the openModal action. Currently, `openModal` store is just a `ModalType`, so I changed the store to have `ModalType` AND some modalData. 

This was needed so that the `DependencyInfo` component could send data on which links to redirect to in the `PrivacyModal` component, and should be useful for other modals in the future too

Default state:
modaldefault

With an open modal:
privacymodal

Expected UI:
image

**How to test**
- [x] Check that privacy modal opens when dependency info warning is clicked
- [x] Check that the links in the privacy modal are correct (privacy statement and download link should correspond to the dependency needed for the framework)
- [x] Check that functions and cosmos modal still open and close properly
- [x] Check that post gen modal still open properly


SahilTara:
I feel like the third party website implies that it is non-Microsoft, since it's a microsoft product? What do you think @jsondoo?

jsondoo:
> I feel like the third party website implies that it is non-Microsoft, since it's a microsoft product? What do you think @jsondoo?

@SahilTara I'm not too sure what you mean by this, what do you suggest for the message?

SahilTara:
> > I feel like the third party website implies that it is non-Microsoft, since it's a microsoft product? What do you think @jsondoo?
> 
> @SahilTara I'm not too sure what you mean by this, what do you suggest for the message?

@jsondoo probably something like "You will now be redirected to a third party website" or something of the sort. The non-microsoft service thing seems implied is what im trying to say

jsondoo:
> > > I feel like the third party website implies that it is non-Microsoft, since it's a microsoft product? What do you think @jsondoo?
> > 
> > 
> > @SahilTara I'm not too sure what you mean by this, what do you suggest for the message?
> 
> @jsondoo probably something like "You will now be redirected to a third party website" or something of the sort. The non-microsoft service thing seems implied is what im trying to say

@SahilTara I talked to Christina as well and we both agreed that a link would be more clear, so will do that

image

src/client/src/containers/PrivacyModal/index.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const { dependency, intl } = props;
+
+  if (dependency === null || dependency === undefined) {
+    return null;
+  }
+
+  const { downloadLink, privacyStatementLink } = dependency;
+
+  const cancelKeyDownHandler = (event: any) => {
+    if (event.keyCode === 13 || event.keyCode === 32) {
```
SahilTara, line 60: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode If you read this it states the keyCode api is deprecated. Try using https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key instead, as its not deprecated and it's more readable. (More docs: https://www.w3schools.com/jsref/event_key_key.asp)

Also you should change the type of the event from any to a KeyboardEvent.

jsondoo: Hey, I think it would be better to address this with a separate PR because we are using keyCode api for our other key handlers. I updated the type of event though

SahilTara: Yeah you're right. Hopefully this will be addressed soon.

streetcoder123:
@jsondoo A couple of comments:

image
- As a user, if I see a url on the pop up, I feel like clicking it but here it is just a text. Should we reconsider this?

- On the pop up, if I click `OK`, it would take me to the appropriate site. But when I go back to the wizard, the pop up is still there. If I click `OK` again, it redirects me again. The only way to get back to the normal flow in wizard is to `X` the pop up. Do you think it would be more user friendly if we close the pop up once the user has clicked `OK` to visit the download site?

src/client/src/actions/modalActions/modalActions.ts
```diff
@@ -22,25 +22,52 @@ const closeModalAction = (): ICloseModal => ({
...
+    dispatch(
+      openModalAction({
+        modalType: MODAL_TYPES.POST_GEN_MODAL,
+        modalData: null
+      })
+    );
+  };
+};
+
+const openPrivacyModalAction = (data: any) => {
```
streetcoder123, line 56: Should we have explicit type for the data field?

src/client/src/actions/modalActions/typeKeys.ts
```diff
@@ -13,4 +14,10 @@ export type ModalType =
   | null
   | MODAL_TYPES.COSMOS_DB_MODAL
   | MODAL_TYPES.AZURE_FUNCTIONS_MODAL
-  | MODAL_TYPES.POST_GEN_MODAL;
+  | MODAL_TYPES.POST_GEN_MODAL
+  | MODAL_TYPES.PRIVACY_MODAL;
+
+export interface ModalState {
+  modalType: ModalType;
+  modalData: any;
```
streetcoder123, line 22: Explicit typing here as well?

jsondoo:
@njindal239 

> * As a user, if I see a url on the pop up, I feel like clicking it but here it is just a text. Should we reconsider this?

This is based on how other Microsoft products do it when Stephane showed me examples although I can see how it's confusing. 

**Edit: After more discussion with the team, we settled on saying "Python/Node download link" rather than the actual URL** 

image
> * On the pop up, if I click `OK`, it would take me to the appropriate site. But when I go back to the wizard, the pop up is still there. If I click `OK` again, it redirects me again. The only way to get back to the normal flow in wizard is to `X` the pop up. Do you think it would be more user friendly if we close the pop up once the user has clicked `OK` to visit the download site?

Good idea

streetcoder123 (commented):
There are a few places where I think we could have explicit type declarations for more type safety.

src/client/src/actions/modalActions/typeKeys.ts
jsondoo: The point of this was to be flexible since `modalData` will vary depending on the modal

jsondoo: We could introduce new types for modalData for each of the modals but I don't really see too much value in doing that right now lol

src/client/src/actions/modalActions/modalActions.ts
jsondoo: See below

src/client/src/actions/modalActions/typeKeys.ts
streetcoder123: Yeah, that makes sense.

Tanya0609 (dismissed):

streetcoder123 (dismissed):
Works Well.
Maybe we could consider having more type safety by adding types for each modal in future :)

src/client/src/containers/DependencyInfo/index.tsx
```diff
@@ -92,9 +107,8 @@ class DependencyInfo extends React.Component<Props> {
       : getSvg.getWarningSvg();
 
     return (
-      <a
-        target={"_blank"}
-        href={downloadLink}
+      <div
+        onClick={() => openPrivacyModal(dependency)}
```
piotrmark, line 112: Please, use role="button" or use button instead of div element
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role

src/client/src/containers/PrivacyModal/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+    id: "privacyModal.privacyStatement",
+    defaultMessage: "Privacy Statement"
+  },
+  OK: {
+    id: "privacyModal.OK",
+    defaultMessage: "OK"
+  }
+});
+
+const PrivacyModal = (props: Props) => {
```
piotrmark, line 50: It's a great candidate for unit tests (just sayin') ;)

src/client/src/containers/PrivacyModal/index.tsx
```diff
@@ -0,0 +1,127 @@
...
+  };
+
+  return (
+    <div>
+      <div className={styles.headerContainer}>
+        <div className={styles.title}>
+          {intl.formatMessage(messages.beingRedirected)}
+        </div>
+        <Cancel
+          tabIndex={0}
```
piotrmark, line 74: Does tabIndex=0 give us something here?

jsondoo: Yes, it lets you navigate using tab onto the X button on the top right

src/client/src/containers/PrivacyModal/index.tsx
jsondoo: coming soon ;)

piotrmark (approved):

streetcoder123 (approved):

Tanya0609 (approved):

PR #822, opened 2019-07-03
Next Button Should not be displayed on Wizard 'Review and Generate' Page

### LINKED ISSUE #815: Footer - 'Next' Button should not be visible on the 'Review and Generate' Page
`Next` button should not be visible on Review and Generate page since it is not required there and causes distraction.

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/30132

**Change**
- We do not want next button to display on the 'review and generate' page of wizard since it is not required.

**Before**
image

**After**
image

closes #815


src/client/src/containers/Footer/index.tsx
```diff
@@ -202,30 +202,28 @@ class Footer extends React.Component<Props> {
...
-                    ? ROUTES.REVIEW_AND_GENERATE
-                    : pathsNext[pathname]
-                }
-              >
-                <FormattedMessage id="footer.next" defaultMessage="Next" />
-              </Link>
+              { pathname !== ROUTES.REVIEW_AND_GENERATE &&
+                <Link
+                  tabIndex={
+                    !isValidNameAndProjectPath ? -1 : 0
```
piotrmark: NIT: feels to me like this negation is not necessary here and it would be more readable:  `isValidNameAndProjectPath ? 0 : -1`

src/client/src/containers/Footer/index.tsx
```diff
@@ -202,30 +202,28 @@ class Footer extends React.Component<Props> {
...
+                  className={classnames(styles.button, buttonStyles.buttonHighlightedBorder, {
+                    [buttonStyles.buttonDark]: !isValidNameAndProjectPath,
+                    [styles.disabledOverlay]:
+                      !isValidNameAndProjectPath
+                  })}
+                  onClick={event => {
+                    this.handleLinkClick(event, pathname);
+                  }}
+                  to={
+                    pathname === ROUTES.REVIEW_AND_GENERATE
```
piotrmark: This is never true, right?

src/client/src/containers/Footer/index.tsx
```diff
@@ -202,30 +202,28 @@ class Footer extends React.Component<Props> {
...
-                  !isValidNameAndProjectPath || this.isReviewAndGenerate()
-                    ? -1
-                    : 0
-                }
-                className={classnames(styles.button, {
-                  [buttonStyles.buttonDark]:
-                    this.isReviewAndGenerate() || !isValidNameAndProjectPath,
-                  [buttonStyles.buttonHighlightedBorder]: !this.isReviewAndGenerate(),
-                  [styles.disabledOverlay]:
-                    !isValidNameAndProjectPath || this.isReviewAndGenerate()
```
piotrmark, line 216: Is `isReviewAndGenerate` function still used somewhere? If not, it should be removed

piotrmark (changes requested):

src/client/src/containers/Footer/index.tsx
streetcoder123: Yeah, it was actually not necessary earlier as well as the button was disabled in the 'Review and Generate' page. I will remove it now.

src/client/src/containers/Footer/index.tsx
streetcoder123: Yes, it is still being used in the Review and Generate button.

annielovesjs (dismissed):
nice! looks good

piotrmark (dismissed):

src/client/src/containers/Footer/index.tsx
jcoc611-microsoft: This is still missing a change right?

jcoc611-microsoft (changes requested):

streetcoder123:
@jcoc611-microsoft This has already been fixed. There seems to a merge conflict now though. I shall resolve that today.

SahilTara (approved):

annielovesjs (approved):

jcoc611-microsoft (approved):

PR #824, opened 2019-07-04
Create resource group for Azure Functions

jasmingoh:
Related PBI: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/29740

The purpose of this PR is to create resource group for user if they choose to put the Azure Functions they create in a new resource group. 

Client side changes will be in another PR (the client will be sending an empty string for `functions.resourceGroup` field in the payload when user choose a new resource). 

So, to test for this PR:
- Add this line `payload.functions.resourceGroup = "";` before line 92 in `generationExperience.ts`. This will force the function that deploys creation of resource group to be kicked off. 
- Launch the wizard
- Take note of the project name you entered in the first page of the wizard. 
- Add Azure Functions when creating your project (the resource group you choose in the form does not matter)
- Go onto Azure portal and check if a new resource group has been created. The name of the new resource group should start with the project name your entered. The Azure functions you created should be in that resource group.

**Note**: Logics that handles creation of resource group for Cosmos (which may be the same as the resource group for Functions) will be handled in another PR. The arm template required for resource group in the generated project will also be addressed in another PR.


src/extension/src/azure/azureServices.ts
```diff
@@ -281,6 +283,24 @@ export class AzureServices extends WizardServant {
...
+    let functionSubscription  = payload.functions.subscription;
+
+    await AzureServices.updateFunctionSubscriptionItemCache(functionSubscription);
+
+    let generatedName = await AzureServices.AzureResourceGroupProvider.generateValidResourceGroupName(projectName, this.usersFunctionSubscriptionItemCache);
+    
+    let resourceGroupSelection: ResourceGroupSelection = {
+      subscriptionItem: AzureServices.usersFunctionSubscriptionItemCache,
+      resourceGroupName: generatedName,
+      location: CONSTANTS.AZURE_LOCATION.CENTRAL_US
```
SahilTara, line 306: why are we forcing Central US?

src/extension/src/azure/azure-resource-group/resourceGroupModule.ts
```diff
@@ -0,0 +1,98 @@
...
+            this.azureResourceClient.subscriptionId !== userSubscriptionItem.subscriptionId
+            ) {
+                this.azureResourceClient = new ResourceManager().getResourceManagementClient(userSubscriptionItem);
+            }
+    }
+
+    public async createResourceGroup(
+        resourceGroupSelection: ResourceGroupSelection
+    ): Promise<string | undefined> {
+        let userSubscriptionItem: SubscriptionItem = resourceGroupSelection.subscriptionItem;
```
SahilTara: Can we destructure instead?

src/extension/src/azure/azure-resource-group/resourceGroupModule.ts
```diff
@@ -0,0 +1,98 @@
...
+    ): Promise<string> {
+        let tries = 0;
+        this.setAzureResourceClient(userSubscriptionItem);
+        let generatedName = this.generateResourceGroupName(name);
+        let isValid = await this.validateResourceGroupName(generatedName);
+
+        while (tries < VALIDATION_LIMIT && !isValid) {
+            generatedName = this.generateResourceGroupName(name);
+            isValid = await this.validateResourceGroupName(generatedName);
+            tries++;
```
SahilTara: What does it mean if tries exceeds validation limit? Does it mean Azure is down or network connectivity problem, or do you think it does not matter to the user?

src/extension/src/generationExperience.ts
```diff
@@ -87,20 +89,19 @@ export class GenerationExperience extends WizardServant {
       payload: { outputPath: enginePayload.path }
     });
 
-    if (payload.selectedFunctions) {
-      serviceQueue.push(
+    if ((payload.selectedFunctions && payload.functions.resourceGroup === "")) {
+      // Reason for this queue is for deploying more than 1 resource group
+      resourceGroupQueue.push(
```
SahilTara, line 94: Is this gonna be used more so when cosmos is also added?

jasmingoh: Yes, this is going to be used more when Cosmos (and App Service) is added

src/extension/src/azure/azureServices.ts
jasmingoh: It was a team (+ coaches) decision to create the resource group in Central US. 

The reason for this feature is to add App Service to the extension, and we are going to use the VS Code App Service extension. The way that their extension does it is that they deploy a resource group for their user, and they default the location to Central US. 

Their extension has an advance mode where user can select the location for their services, but this feature is targetted towards our users, i.e. beginner/intermediate developers, who won't really care about location of the resource group.

Our assumption is that, if a developer cares enough about the location of their resource group, they are probably already familiar with Azure, and will have no problem with creating one using the portal instead. In that case, they will just go through the existing flow (i.e. choose resource group from the drop down)

src/extension/src/azure/azure-resource-group/resourceGroupModule.ts
jasmingoh: What do you mean?

src/extension/src/azure/azure-resource-group/resourceGroupModule.ts
jasmingoh: It is very unlikely that tries will exceed the validation limit because a new name (with a new timestamp) will be generated before it will be validated again. 

Just to address your question, I tried disconnecting my computer, and I was not able to add any Azure resources, which means that this method will never be trigerred if the user has connectivity problem.

So, if there is ever a case that tries exceed the validation limit, it is more likely that Azure is down. 

Please let me know if there is any other edge cases I should consider.

src/extension/src/azure/azure-resource-group/resourceGroupModule.ts
SahilTara:
```ts
const {userSubscriptionItem, resourceGroupName, location} = resourceGroupSelection;
```

src/extension/src/azure/azure-resource-group/resourceGroupModule.ts
SahilTara: There is another edge case where the user signs out of their azure account using the command pallet command rather than through the wizard after adding azure. But this is a known issue which can't be fixed yet, or at least can't be fixed nicely.  There is a PR up for the API solution that would help us, but the vscode team is still doing tests on it to see if there are any performance implications.  (I'm following this issue really closely).

annielovesjs:
Hmm, it didn't deploy for me: 
functionsDeployFailed

jasmingoh:
@annielovesjs I just looked at the logs on Azure portal. Here's the [link](https://ms.portal.azure.com/#@microsoft.onmicrosoft.com/resource/subscriptions/31add260-f4a3-497d-8b39-e35101c87f55/resourceGroups/asdas_testjasmin_20197418326/deployments).
image

It seems like we can't create Azure Functions in Canada East at all. I tried manually adding Azure Functions to the same resource group and set the location to Canada East, and it failed too. But I was able to manually add Azure Functions to West US in that resource group.

I also tested the extension version that we have on dev and Azure Functions deployment also fails when we choose Canada East as the location.

@piotrmark @jcoc611-microsoft @SahilTara Any idea what we can do about this?

jasmingoh:
@annielovesjs Can you try other locations when choosing Azure function? Thanks!

annielovesjs:
> @annielovesjs Can you try other locations when choosing Azure function? Thanks!

In that case, we probably shouldnt have those as an option for the dropdown. Will check to see if more locations fail as well

annielovesjs:
@jasmingoh the locations it won't work for are in the UAE, Africa, US GOV + US DoD (except for Arizona), Korea south, France, UK West, Germany, West India + Central India, Australia central  + Australia central 2, Canada East

annielovesjs (dismissed):
@jasmingoh on the locations that do work, it does get deployed properly - but we should properly make a separately pr to address the location problem

SahilTara:
Yeah we knew about this problem, its a bug in the azure sdk actually, it technically shouldn't even be showing those locations due to there being a capacity problem or something of the sort. There's a known issue for this, can't find it right now on one of the azure githubs or something. According to danish the best thing you can do for now is to hard code the values that don't work out of the returned info. Also apparently most of the locations that break if not all are "Internal to microsoft only", so try testing with your personal emails.

src/extension/src/azure/azure-resource-group/resourceGroupModule.ts
```diff
@@ -0,0 +1,96 @@
...
+            throw new AuthorizationError(CONSTANTS.ERRORS.AZURE_RESOURCE_CLIENT_NOT_DEFINED);
+        }
+        let exist = await this.azureResourceClient.resourceGroups.checkExistence(name);
+        return await !exist;
+    }
+
+    private generateResourceGroupName(userProjectName: string): string {
+        let unixTimestamp = Date.now();
+        let suffix = this.unixToSuffix(unixTimestamp);
+        return userProjectName + "_" + suffix;
```
SahilTara: Apparently the max length of a resource group is 90 characters. 
image

What happens if the project name is > 81 characters? 

Look at this snippit of the library code for more info:

https://github.com/Azure/azure-sdk-for-node/blob/40bad8fd52c0cc2264a0900a4d412c8f68aef263/lib/services/resourceManagement/lib/resource/operations/resourceGroups.js#L40-L80

jasmingoh: Thanks @SahilTara, totally did not think about this. I'll find a way to handle this after talking to the team.

SahilTara: My suggestion would be having a max project name length of 75 if you want to keep that suffix. (90 - 14 [character length of 20191212121212] -1 [rounds the number out])

jasmingoh: Exactly our thoughts, the change is now in PR (https://github.com/microsoft/WebTemplateStudio/pull/836)

src/extension/src/azure/azure-resource-group/resourceGroupModule.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    private unixToSuffix(unixTimestamp: any): string {
+        let fullDate = new Date(unixTimestamp);
+        let year  = fullDate.getFullYear().toString();
+        // getMonth() returns month as a zaro-based value
```
jsondoo: @jasmingoh typo here

SahilTara: Wow it actually took me 5 minutes staring at the screen to see the typo, good catch.

jasmingoh: Thanks @jsondoo!

SahilTara (dismissed):
Think this is good to go

src/extension/src/azure/azure-resource-group/resourceGroupModule.ts
```diff
@@ -0,0 +1,96 @@
...
+                this.azureResourceClient = new ResourceManager().getResourceManagementClient(userSubscriptionItem);
+            }
+    }
+
+    public async createResourceGroup(
+        resourceGroupSelection: ResourceGroupSelection
+    ): Promise<string | undefined> {
+        const {subscriptionItem, resourceGroupName, location} = resourceGroupSelection;
+
+        let azureResourceClient: ResourceManagementClient = new ResourceManager().getResourceManagementClient(subscriptionItem);
```
piotrmark: This local variable hides class property with the same name

src/extension/src/azure/azure-resource-group/resourceGroupModule.ts
```diff
@@ -0,0 +1,96 @@
...
+
+        let options: ResourceGroup = {
```
piotrmark: When we don't modify a variable after first assignment, we should use const instead of let.

src/extension/src/azure/azure-resource-group/resourceGroupModule.ts
```diff
@@ -0,0 +1,96 @@
...
+        let year  = fullDate.getFullYear().toString();
+        // getMonth() returns month as a zero-based value
+        let month = (fullDate.getMonth() + 1).toString();
+        let date  = fullDate.getDate().toString();
+        let hour  = fullDate.getHours().toString();
+        let min   =
+         fullDate.getMinutes().toString();
+        let sec   = fullDate.getSeconds().toString();
+        return year.concat(month, date, hour, min, sec);
+
```
piotrmark, line 100: Was this file Prettified? :)

src/extension/src/azure/azure-resource-group/resourceGroupModule.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    private async validateResourceGroupName(
+        name: string
+    ): Promise<boolean> {
+        if (this.azureResourceClient === undefined) {
+            throw new AuthorizationError(CONSTANTS.ERRORS.AZURE_RESOURCE_CLIENT_NOT_DEFINED);
+        }
+        let exist = await this.azureResourceClient.resourceGroups.checkExistence(name);
+        return await !exist;
```
piotrmark: Why do we need 2nd await here?

SahilTara: Yeah it does seem odd to me as well.

src/extension/src/azure/azure-resource-group/resourceGroupModule.ts
jasmingoh: oops.. just pushed the prettified version :)

piotrmark (dismissed):
Actually, my comment about const vs let can be applied to majority of let declarations here :)
Other than that, it's good to go :)

piotrmark (approved):

jsondoo (approved):

SahilTara (approved):

annielovesjs (approved):

PR #825, opened 2019-07-04
Azure Functions Form Update - Part 1

jsondoo:
**Changes**
- additional sublabels for the form (finalized the strings with the team)
- scrollbar for the long form
- more reponsiveness
- runtime stack is defaulted since there's only one option

Most of the changes are reordering of the sections

**Expected**
expected


src/client/src/containers/AzureFunctionsModal/index.tsx
```diff
@@ -310,17 +304,18 @@ const AzureFunctionsResourceModal = (props: Props) => {
...
-          <div>{leftHeader}</div>
+          <div className={styles.leftHeader}>{leftHeader}</div>
           {links[formSectionId] && (
             <a
               tabIndex={disabled ? -1 : 0}
               className={styles.link}
               href={links[formSectionId]}
             >
-              {props.intl.formatMessage(messages.createNew)}
+              {rightHeader}
```
jsondoo, line 314: This was a bug with the original code, `rightHeader` wasn't being used at all but it's supposed to be here I think

src/client/src/mockData/mockVsCodeApi.ts
```diff
@@ -84,7 +84,20 @@ const mockVsCodeApi = () => ({
...
+                  { label: "ResourceGroupMock2", value: "ResourceGroupMock2" },
+                  { label: "ResourceGroupMock3", value: "ResourceGroupMock3" },
+                  { label: "ResourceGroupMock4", value: "ResourceGroupMock4" },
+                  { label: "ResourceGroupMock5", value: "ResourceGroupMock5" },
+                  { label: "ResourceGroupMock6", value: "ResourceGroupMock6" },
+                  { label: "ResourceGroupMock7", value: "ResourceGroupMock7" },
+                  { label: "ResourceGroupMock8", value: "ResourceGroupMock8" },
+                  { label: "ResourceGroupMock9", value: "ResourceGroupMock9" },
+                  {
+                    label: "ResourceGroupMock10",
```
jasmingoh: Format does not look consistent with the rest

src/client/src/mockData/mockVsCodeApi.ts
```diff
@@ -84,7 +84,20 @@ const mockVsCodeApi = () => ({
                 locations: [{ label: WEST_US, value: WEST_US }],
                 resourceGroups: [
                   { label: RESOURCE_GROUP_MOCK, value: RESOURCE_GROUP_MOCK },
```
jasmingoh: should this label and value be changed so that it is consistent with the rest?

src/client/src/mockData/mockVsCodeApi.ts
```diff
@@ -100,7 +113,20 @@ const mockVsCodeApi = () => ({
               payload: {
                 locations: [{ label: WEST_US, value: WEST_US }],
                 resourceGroups: [
-                  { label: RESOURCE_GROUP_MOCK, value: RESOURCE_GROUP_MOCK }
+                  { label: RESOURCE_GROUP_MOCK, value: RESOURCE_GROUP_MOCK },
```
jasmingoh: same comment for consistent label and value as above

src/client/src/mockData/mockVsCodeApi.ts
```diff
@@ -100,7 +113,20 @@ const mockVsCodeApi = () => ({
...
+                  { label: "ResourceGroupMock2", value: "ResourceGroupMock2" },
+                  { label: "ResourceGroupMock3", value: "ResourceGroupMock3" },
+                  { label: "ResourceGroupMock4", value: "ResourceGroupMock4" },
+                  { label: "ResourceGroupMock5", value: "ResourceGroupMock5" },
+                  { label: "ResourceGroupMock6", value: "ResourceGroupMock6" },
+                  { label: "ResourceGroupMock7", value: "ResourceGroupMock7" },
+                  { label: "ResourceGroupMock8", value: "ResourceGroupMock8" },
+                  { label: "ResourceGroupMock9", value: "ResourceGroupMock9" },
+                  {
+                    label: "ResourceGroupMock10",
```
jasmingoh: same as comment for formatting as above

src/client/src/mockData/mockVsCodeApi.ts
jsondoo: So for this one, we could just change `RESOURCE_GROUP_MOCK` to hardcoded string, or we could create more const variables at the top like here:
https://github.com/microsoft/WebTemplateStudio/blob/297d2537b4328eaedb9e5913228d4fed14caf6a8/src/client/src/mockData/mockVsCodeApi.ts#L8

I will use const

src/client/src/mockData/mockVsCodeApi.ts
jsondoo: Yea, this was done by Prettier not me lol

jsondoo: Maybe I should stop using Prettier everywhere

src/client/src/mockData/mockVsCodeApi.ts
```diff
@@ -84,7 +94,46 @@ const mockVsCodeApi = () => ({
...
+                    value: RESOURCE_GROUP_MOCK_9
+                  },
+                  {
+                    label: RESOURCE_GROUP_MOCK_10,
+                    value: RESOURCE_GROUP_MOCK_10
+                  },
+                  {
+                    label: RESOURCE_GROUP_MOCK_11,
+                    value: RESOURCE_GROUP_MOCK_11
+                  }
```
streetcoder123: What do all these mocks mean?

jsondoo: I added this to test the longer dropdowns since we set a limit on how long the dropdown is in the css

SahilTara: Why not just use a loop to generate this?

jsondoo: I could, but I don't think it's necessary. Do you guys have any thoughts on this? @njindal239 @jasmingoh

jasmingoh: Agree with @jsondoo, I don't think it is necessary to create a loop and it doesn't really make sense here. Maybe we can just use the same constant for all labels and values since this is just a mock?

SahilTara: Yeah using the same value works, its just a lot of unnecessary noise right now I feel. Any way you can reduce it is good. The map/loop is just the first thing that came to mind due to the repeated elements with only a number being different. 

However I can see the loop being a bit less readable as well. I thought about it a bit more, and using Jasmin's suggestion and mine I thought of the following:

```ts
const mockResourceGroups = Array.from({length: 12}).fill({label: RESOURCE_GROUP_MOCK, value: RESOURCE_GROUP_MOCK});
```

Then later on in the code you could do 

```ts
  resourceGroups: mockResourceGroups
```
for both functions and cosmos. You could also get rid of all the constants. Only do it if you think its more readable. Otherwise I think @jasmingoh 's idea of simply repeating RESOURCE_GROUP_MOCK is a better idea than the current solution.

src/client/src/mockData/mockVsCodeApi.ts
piotrmark: No, you should have consistent settings and use it everywhere. Otherwise we will have a lot of noise (formatting changes) in PRs

src/client/src/mockData/mockVsCodeApi.ts
jsondoo: @SahilTara Yea, I like this solution the most

src/client/src/containers/AzureFunctionsModal/index.tsx
SahilTara: Nice catch

SahilTara (approved):
Looks good for the most part, nice addition of comments, helps with the readability of things.

xtinah-w (approved):
Looks good and works on my system

streetcoder123 (approved):

jasmingoh (approved):

annielovesjs (approved):

PR #826, opened 2019-07-04
Create state for 'Create Project' Button

### LINKED ISSUE #816: Create state for 'Create Template' Button
We need 'Create Template' Button not to be visible until the user reaches the 'Review and Generate' page. However, once the user reaches that page, it should be visible no matter where the user navigates to.

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/30133

**Expected Behavior**
'Create Project' Button is not visible until the user reaches the 'Review and Generate' page. However, once the user reaches that page, it remains visible no matter where the user navigates to.

closes #816


src/client/src/actions/wizardInfoActions/typeKeys.ts
```diff
@@ -5,5 +5,6 @@ export enum WIZARD_INFO_TYPEKEYS {
   UPDATE_TEMPLATE_GENERATION_STATUS_MESSAGE = "WTS/postgen/UPDATE_TEMPLATE_GENERATION_STATUS_MESSAGE",
   UPDATE_TEMPLATE_GENERATION_STATUS = "WTS/postgen/UPDATE_TEMPLATE_GENERATION_STATUS",
   RESET_WIZARD = "RESET_WIZARD",
-  UPDATE_DEPENDENCY_INFO = "WTS/dependency/UPDATE_DEPENDENCY_INFO"
+  UPDATE_DEPENDENCY_INFO = "WTS/dependency/UPDATE_DEPENDENCY_INFO",
+  UPDATE_CREATE_TEMPLATE_BUTTON = "UPDATE_CREATE_TEMPLATE_BUTTON"
```
SahilTara: isn't it CREATE_PROJECT_BUTTON Now? Many things have to follow suit as well (IUpdateCreateTemplateButton and others)

src/client/src/actions/wizardInfoActions/updateCreateTemplateButton.ts
```diff
@@ -0,0 +1,13 @@
+import {WIZARD_INFO_TYPEKEYS} from './typeKeys'
```
SahilTara: Think you should apply prettier to this file.

src/client/src/containers/Footer/index.tsx
```diff
@@ -273,6 +277,9 @@ const mapDispatchToProps = (
   },
   openPostGenModal: () => {
     dispatch(openPostGenModalAction());
+  },
+  updateCreateTemplateButton: () => {
+    dispatch(updateCreateTemplateButtonAction(true));
```
SahilTara: I don't think there's a point of this action accepting a boolean if it only ever accepts "true".
Would probably change your code to something like this, 
```ts
    switch(action.type) {
        case WIZARD_INFO_TYPEKEYS.UPDATE_CREATE_TEMPLATE_BUTTON:
            return true;
        case WIZARD_INFO_TYPEKEYS.RESET_WIZARD:
            return false;
        default:
            return state;
    }
```

streetcoder123: I agree here. However, I feel like keeping the reducer to update whatever boolean value comes with the action.payload is more generic to allow calling the reducer with `false` as the new state. 
That is why, I kept the name to `UPDATE_CREATE_TEMPLATE_BUTTON` rather than `ENABLE_CREATE_TEMPLATE_BUTTON` as this could either flip the state to `true` or `false` depending upon the `action.payload`

SahilTara: do you ever plan on making it false again other than after you reset the wizard using that action? Or can you see a potential use case where you would want to?

streetcoder123: I cannot think of a potential situation right now. Just want to have it since it feels to be more generic.

SahilTara: Cool, its not a big deal so i'm fine with it.

SahilTara (dismissed):

SahilTara (dismissed):

SahilTara (dismissed):

src/client/src/containers/Footer/index.tsx
```diff
@@ -268,6 +279,9 @@ const mapDispatchToProps = (
   },
   openPostGenModal: () => {
     dispatch(openPostGenModalAction());
+  },
+  updateCreateProjectButton: () => {
```
piotrmark: NIT: Maybe `enableCreateProjectButton` would be a better name here as we don't pass any argument and imply true in this function

piotrmark (dismissed):

src/client/src/actions/wizardInfoActions/typeKeys.ts
```diff
@@ -5,5 +5,6 @@ export enum WIZARD_INFO_TYPEKEYS {
   UPDATE_TEMPLATE_GENERATION_STATUS_MESSAGE = "WTS/postgen/UPDATE_TEMPLATE_GENERATION_STATUS_MESSAGE",
   UPDATE_TEMPLATE_GENERATION_STATUS = "WTS/postgen/UPDATE_TEMPLATE_GENERATION_STATUS",
   RESET_WIZARD = "RESET_WIZARD",
-  UPDATE_DEPENDENCY_INFO = "WTS/dependency/UPDATE_DEPENDENCY_INFO"
+  UPDATE_DEPENDENCY_INFO = "WTS/dependency/UPDATE_DEPENDENCY_INFO",
+  UPDATE_CREATE_PROJECT_BUTTON = "UPDATE_CREATE_TEMPLATE_BUTTON"
```
jsondoo: should change the string to "UPDATE_CREATE_PROJECT_BUTTON" as well

src/client/src/containers/Footer/index.tsx
streetcoder123: I changed it to accept a boolean argument to update the state so that we can use the same action now to make it `false` as well. :)

SahilTara (approved):

jasmingoh (approved):
Works for me

annielovesjs (approved):

PR #828, opened 2019-07-04
client calls extension instead of CoreTS to get frameworks

Tanya0609:
Task:https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/WTS%20Team/Intern%20GitHub/Y19-VS/07

(this PR did not address extension calls CoreTS CLI, instead of call CoreTS API)
Ideally, client should call extension via message passing, rather than call CoreTS api directly. There are few API calls to CoreTS in our client right now. This PR is to address the **getFrameworks** API call to CoreTS.  Thus, in my PR: 
- from client, it sends getFrameworks command to extension
- in clinet, removed thunk action (since we don't need to resolve the promise from an api call then disptach update the store action)
- in extension, I added a module to listen and return response to getFrameworks command
- in client App.tsx, it listens to response(frameworks data) from extension and update the store with the response

**How to test:**
1. start the extension (once with preview mode, and the other time without preview mode)
2. the select frameworks page should still be the same
3. the extensions should still be able to generate templates

frameworks


Tanya0609:
closed due to naming

src/client/src/utils/constants.ts
```diff
@@ -78,7 +80,8 @@ const EXTENSION_MODULES = {
   TELEMETRY: "Telemetry",
   VALIDATOR: "Validator",
   VSCODEUI: "VSCodeUI",
-  DEPENDENCYCHECKER: "DependencyChecker"
+  DEPENDENCYCHECKER: "DependencyChecker",
+  CoreTSModule: "CoreTSModule"
```
jcoc611-microsoft: NIT: For consistency, should it be `CORETS` instead of `CoreTsModule`?

src/client/src/App.tsx
```diff
@@ -125,6 +133,23 @@ class App extends React.Component<Props> {
     window.addEventListener("message", event => {
       const message = event.data;
       switch (message.command) {
+        // get frameworks from extension message
+        case EXTENSION_COMMANDS.GET_FRAMEWORKS:
+          this.props.getFrontendFrameworksSuccess(
+            getFrameworks(
```
jcoc611-microsoft: I think it might make more sense to have this function be called something like `parseFrameworksPayload`. It might be confusing if it is still called `getFrameworks` but it doesn't retrieve frameworks anymore

src/client/src/actions/wizardContentActions/getFrameworks.ts
```diff
@@ -1,33 +1,17 @@
...
 type FrameworkType = "frontend" | "backend";
 
-// thunk
-export const getFrameworks = async (
-  projectType: string,
+export const getFrameworks = (
+  frameworksJson: any,
   type: FrameworkType,
-  isPreview: boolean,
-  serverPort: number
```
jcoc611-microsoft, line 13: I think we should also remove all the logic for `serverPort`. This includes the `EXTENSION_COMMANDS.GET_PORT` command, both from the client side and the extension side, as well as the redux state that has the server port. This is because server port is no longer used in client.

src/extension/src/CoreTSModule.ts
```diff
@@ -0,0 +1,34 @@
...
+export class CoreTSModule extends WizardServant {
+    clientCommandMap: Map<
+        ExtensionCommand,
+        (message: any) => Promise<IPayloadResponse>
+    >;
+
+    constructor() {
+        super();
+        this.clientCommandMap = this.defineCommandMap();
+        console.log("creating CoreTSmodule");
```
jcoc611-microsoft: Please remove

src/client/src/containers/SelectFrameworks/index.tsx
```diff
@@ -22,6 +26,17 @@ type Props = ISelectFrameworksProps;
 class SelectFrameworks extends React.Component<Props> {
   componentDidMount() {
     const { vscode, isPreview } = this.props;
+    // send extension commands to load frameworks
+    vscode.postMessage({
+      module: EXTENSION_MODULES.CoreTSModule,
+      command: EXTENSION_COMMANDS.GET_FRAMEWORKS,
+      track: false,
+      text: "Sending get frameworks commands...",
```
jcoc611-microsoft: I think this one should be removed?

src/client/src/utils/constants.ts
Tanya0609: Good Catch, naming should be capitalized. Will update that in my PR.

Also, I was thinking CoreTS, however, your refactor file has a similar name "coreTemplateStudio.ts", this was why I named it CoreTSModule. What do you think? Should I stick with "CORETSMODULE" or change to "CORETS"?

Tanya0609: PR updated

src/client/src/containers/SelectFrontEndFramework/index.tsx
```diff
@@ -56,14 +50,7 @@ const messages = defineMessages({
...
   public componentDidMount() {
-    const { getFrontendFrameworks, isPreview, serverPort } = this.props;
-    if (getFrontendFrameworks) {
-      getFrontendFrameworks(
-        WIZARD_CONTENT_INTERNAL_NAMES.FULL_STACK_APP,
-        isPreview,
-        serverPort
-      );
-    }
+    const { isPreview, serverPort } = this.props;
```
SahilTara: Why is the componentDidMount still here? Isn't it empty or am i missing something?

src/client/src/actions/wizardContentActions/getFrameworks.ts
```diff
@@ -1,38 +1,20 @@
-import EngineAPIService from "../../services/EngineAPIService";
 import getSvgUrl from "../../utils/getSvgUrl";
 import { IMetadata } from "../../types/metadata";
 import { IOption } from "../../types/option";
 
-type FrameworkType = "frontend" | "backend";
```
SahilTara, line 6: Why did we remove this type?

src/client/src/containers/SelectBackendFramework/index.tsx
```diff
@@ -44,15 +38,9 @@ const messages = defineMessages({
...
   public componentDidMount() {
-    const { getBackendFrameworks, isPreview, serverPort } = this.props;
-    if (getBackendFrameworks !== undefined) {
-      getBackendFrameworks(
-        WIZARD_CONTENT_INTERNAL_NAMES.FULL_STACK_APP,
-        isPreview,
-        serverPort
-      );
-    }
+    const { isPreview, vscode } = this.props;
```
SahilTara: do we need this componentDidMount?

src/client/src/containers/SelectBackendFramework/index.tsx
```diff
@@ -3,34 +3,28 @@ import { connect } from "react-redux";
...
-    serverPort: number
-  ) => void;
 }
 
 interface ISelectBackendProps {
   options: IOption[];
   selectedBackend: ISelected;
   serverPort: number;
   isPreview: boolean;
+  vscode: any;
```
SahilTara: Think we should be typesafe here.

src/client/src/containers/SelectBackendFramework/index.tsx
Tanya0609: Good Catch, since we are not calling API from here, it is not used anymore. will update my PR.

src/client/src/containers/SelectBackendFramework/index.tsx
Tanya0609: will update in my PR

src/client/src/actions/wizardContentActions/getFrameworks.ts
Tanya0609: I was trying to assign constants into this type instead of hardcore the values here. But it gave error. And tried to assign constants to enum, it cannot detect the constants. I did not find a better way of doing it. Do you have any suggestions?

Before this is a better solution, I will just revert and stick with the original way to handle constraining the values allowed for this parameter.

src/extension/src/CoreTSModule.ts
jasmingoh: Do we need this console log or is it for debugging?

Tanya0609: updated PR

jasmingoh (dismissed):

jcoc611-microsoft (changes requested):

src/client/src/actions/wizardContentActions/getFrameworks.ts
SahilTara: why not just create
```
enum FrameworkType {
  FRONTEND = "frontend",
  BACKEND = "backend",
}
```
and use that in the constants file and use it everywhere.

src/client/src/containers/SelectFrameworks/index.tsx
Tanya0609: updated PR

src/client/src/actions/wizardContentActions/getFrameworks.ts
Tanya0609: server port is still in used by getPages, but I agree, once all the api calls are through extension, then we should remove everything related to the old logic

src/client/src/App.tsx
Tanya0609: good suggestion :) I will update it in my PR

src/extension/src/CoreTSModule.ts
Tanya0609: removed :)

src/client/src/actions/wizardContentActions/getFrameworks.ts
Tanya0609: This still feels hard coding and same as what it has in the code. Is there a way to utilize values from CONSTANT file?

SahilTara: I'm saying replace the values in the constants file with the enum, and export it

Tanya0609: updated PR :)

src/extension/src/CoreTSModule.ts
```diff
@@ -0,0 +1,33 @@
+import { WizardServant, IPayloadResponse } from "./wizardServant";
```
jasmingoh: NIT: The first character of the module name should not be capitalized, i.e. coreTSModule.ts

Tanya0609: PR updated :)

jasmingoh (dismissed):

jasmingoh (dismissed):

src/client/src/utils/constants.ts
```diff
@@ -80,7 +85,8 @@ const EXTENSION_MODULES = {
   TELEMETRY: "Telemetry",
   VALIDATOR: "Validator",
   VSCODEUI: "VSCodeUI",
-  DEPENDENCYCHECKER: "DependencyChecker"
+  DEPENDENCYCHECKER: "DependencyChecker",
+  CORETS: "CoreTSModule"
 };
 
 // Define extension commands here that should be received from the extension
```
jcoc611-microsoft, line 92: Should remove `EXTENSION_COMMANDS.GET_PORT`

src/client/src/App.tsx
```diff
@@ -126,6 +133,23 @@ class App extends React.Component<Props> {
...
+            parseFrameworksPayload(
+              message.payload.frameworks,
+              FRAMEWORK_TYPE.BACKEND,
+              message.payload.isPreview
+            )
+          );
+          break;
         case EXTENSION_COMMANDS.GET_DEPENDENCY_INFO:
           this.props.updateDependencyInfo(message.payload);
           break;
```
jcoc611-microsoft, line 155: Should remove the `GET_PORT` case below

src/client/src/actions/wizardContentActions/parseFrameworksPayload.ts
```diff
@@ -1,38 +1,21 @@
...
-      );
-    } else {
-      console.log("FAILED");
-      return [];
-    }
-  } catch (error) {
-    console.log(error);
-    return [];
-  }
+export const parseFrameworksPayload = (
```
streetcoder123: Probably, we should move this file to the utils folder since these functions are not actions but helper functions.

Tanya0609: PR updated

src/client/src/App.tsx
Tanya0609: PR updated

src/client/src/utils/constants.ts
Tanya0609: PR updated :)

jasmingoh (dismissed):

jcoc611-microsoft (dismissed):
Good work!

streetcoder123 (dismissed):

jasmingoh (approved):

jasmingoh:
build is successful!

streetcoder123 (approved):

jcoc611-microsoft (approved):

Issue #817, opened 2019-07-03
keyCode API is deprecated


jsondoo:
> https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode If you read this it states the keyCode api is deprecated. Try using https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key instead, as its not deprecated and it's more readable. (More docs: https://www.w3schools.com/jsref/event_key_key.asp)

Right now we use `KeyboardEvent.keyCode` for all our key handlers, we should look into using `KeyboardEvent.key` instead

PR #829, opened 2019-07-04
Dependency Info Tab Navigation

jsondoo:
This makes the component be tabbable with a keydown handler (you should be able to tab over and then press space or enter to interact with it)


annielovesjs (approved):
works well!

src/client/src/containers/DependencyInfo/index.tsx
```diff
@@ -119,9 +119,17 @@ class DependencyInfo extends React.Component<Props> {
       ? getSvg.getGreenCheckSvg()
       : getSvg.getWarningSvg();
 
+    const keyDownHandler = (event: React.KeyboardEvent<HTMLDivElement>) => {
+      if (event.keyCode === 13 || event.keyCode === 32) {
```
streetcoder123, line 123: Should we have 13 and 32 as constants somewhere so that it be clear for any other developer looking at this code as to what do they refer to?

src/client/src/containers/DependencyInfo/index.tsx
```diff
@@ -119,9 +119,17 @@ class DependencyInfo extends React.Component<Props> {
...
+        openPrivacyModal(dependency);
+      }
+    };
+
     return (
       <div
         role="button"
+        tabIndex={0}
```
streetcoder123, line 131: Just out of curiosity, what does `tabIndex={0}` mean?

src/client/src/containers/DependencyInfo/styles.module.css
```diff
@@ -8,6 +8,10 @@
   text-decoration: none;
 }
 
+.dependencyContainer:focus {
+  outline: 1px solid var(--vscode-contrastActiveBorder);
+}
+
```
streetcoder123, line 14: What does this style do?

src/client/src/containers/DependencyInfo/index.tsx
jsondoo: Yea, I think we can change ALL keyhandlers to be with constants when we resolve #817

src/client/src/containers/DependencyInfo/index.tsx
jsondoo: Good question, it allows you to navigate using tab to the element. The number defines the order of navigation (what is focused first)

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

src/client/src/containers/DependencyInfo/styles.module.css
jsondoo: It adds the orange border when u navigate to it using tab (we use this css in all the navigable components).

streetcoder123 (approved):
Works Well
We need to make sure to have #817 solved soon

piotrmark (approved):

jasmingoh (approved):

PR #830, opened 2019-07-04
Template tests part 2 

annielovesjs:
How it works
- this script automates running yarn install, yarn start and yarn test for all the generated projects from generate-test.js and throwing any errors that may appear 
- Running yarn start on the projects can allow us to detect errors in server side code (thrown when we detect an error that has a traceback)
- Running yarn test on the projects allows us to also the available unit tests for the frontend frameworks (currently only angular has a working unit test)
- After the test finishes (and if an error isn't thrown), all the generated projects from this test will be deleted  

How to test:
**1st iteration** -> this test should pass 
- cd WebTemplateStudio/src/extension 
- run yarn template-tests to generate the templates and run the test

**2nd iteration** -> this test should fail
- go to any angular project and navigate to app.component.ts 
- change the title attribute inside class AppComponent to a different string 
- run yarn template-tests 
- script should throw an error after it finishes running yarn test on that project and terminate

What to expect:
- takes about 10 minutes in total for both generate-test and run-tests to finish running


src/extension/package.json
```diff
@@ -74,6 +74,7 @@
     "postinstall": "node ./node_modules/vscode/bin/install",
     "test": "npm run compile && node ./node_modules/vscode/bin/test",
```
jcoc611-microsoft, line 75: If we aren't using this `npm test` for anything, maybe you could replace it with your tests

```
"test": "yarn generate &&  yarn run-tests"
```

jcoc611-microsoft:
Did you get this to work with Flask? I'm getting the following with a fresh build of this branch:

```
$ python3 server/server.py || py -3 server/server.py || python server/server.py

[1]
Traceback (most recent call last):
[1]   File "server/server.py", line 1, in <module>
[1]     from flask import Flask, jsonify, make_response, send_from_directory
[1] ModuleNotFoundError: No module named 'flask'

D:\other\WTS\WebTemplateStudio\src\extension\out\scripts\run-tests.js:89
                throw new Error("Error found in project");
```

src/extension/src/scripts/run-tests.ts
```diff
@@ -0,0 +1,150 @@
...
+const fs = require("fs");
+
+const testFolder = path.join(__dirname, "..", "..", "src", "template_test");
+console.log(testFolder);
+let files: string[] = [];
+
+//Recursively delete all generated projects in template_test
+let deleteFolderRecursive = function(pathname: string) {
+  let files = [];
+  if (fs.existsSync(pathname)) {
```
jasmingoh: Is it possible to call the script to generate templates if the pathname is not found? That way, we only need to run 1 command, i.e. `yarn run-tests`, to run the tests (instead of `yarn generate` and then `yarn run-tests`)

src/extension/src/scripts/run-tests.ts
```diff
@@ -0,0 +1,150 @@
...
+  try {
+    serverProcess = child_process.exec(
+      "yarn start",
+      {
+        cwd: path.join(testFolder, file, file),
+        stdio: "inherit"
+      },
+      (error: any, stdout: any, stderr: any) => {
+        if (error) {
+          console.error(`Error from running yarn start: ${error}`); //TODO: mark failure
```
jasmingoh: is there more work for this TODO in this PR?

src/extension/package.json
SahilTara: I feel like test could still be useful if we add extension side tests. Not sure though.

src/extension/src/scripts/run-tests.ts
```diff
@@ -0,0 +1,150 @@
+let child_process = require("child_process");
+const path = require("path");
+const fs = require("fs");
+
+const testFolder = path.join(__dirname, "..", "..", "src", "template_test");
+console.log(testFolder);
+let files: string[] = [];
+
+//Recursively delete all generated projects in template_test
+let deleteFolderRecursive = function(pathname: string) {
```
SahilTara: I don't think this method is necessary. 

Our package.json has:
https://github.com/microsoft/WebTemplateStudio/blob/e540335646b1601e26efb120fd76a6c600cf9284/src/extension/package.json#L82

The same thing can be achieved by doing
```ts
const del = require("del");
del.sync(pathname);
```

src/extension/src/scripts/run-tests.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+};
+
+fs.readdirSync(testFolder).forEach((file: string) => {
+  files.push(file.toString());
+});
+
+function kill(pid: any) {
+  try {
+    child_process.execSync("taskkill /PID " + pid + " /T /F", function(
```
SahilTara, line 16: mac and linux?

src/extension/src/scripts/run-tests.ts
```diff
@@ -0,0 +1,150 @@
...
+      if (error !== null) {
+        console.log("Error from kill: " + error);
+      }
+    });
+  } catch (err) {
+    console.log(`Error from taskkill: ${err}`);
+  }
+}
+
+function sleep(ms: number) {
```
SahilTara, line 32: Side note: Feel like we should add this to a utils class or something, see it being used in many places now.

src/extension/src/scripts/run-tests.ts
piotrmark: This test will run just on CI VM, so I don't think we have to handle all the platforms, right?

src/extension/src/scripts/run-tests.ts
```diff
@@ -0,0 +1,150 @@
+let child_process = require("child_process");
```
piotrmark: Can be const?

src/extension/src/scripts/run-tests.ts
piotrmark: Should we throw here?

src/extension/src/scripts/run-tests.ts
```diff
@@ -0,0 +1,150 @@
...
+      (error: any, stdout: any, stderr: any) => {
+        if (error) {
+          console.error(`Error from running yarn start: ${error}`); //TODO: mark failure
+          return;
+        }
+        if (stderr) {
+          throw stderr;
+        }
+        console.log(`Stdout from yarn start: ${stdout}`);
+        console.log(`Stderr from yarn start: ${stderr}`);
```
piotrmark: I think we don't need this line as we throw above if stderr exists

src/extension/src/scripts/run-tests.ts
```diff
@@ -0,0 +1,150 @@
...
+      console.log(data);
+      if (data.toString().indexOf("FAILED") > -1) {
+        throw new Error("Error: Test failed");
+      }
+    });
+  } catch (err) {
+    console.log("Test errored out");
+    throw err;
+  }
+  await sleep(80000);
```
piotrmark, line 120: Is there a reason why we don't run yarn test synchronously (like yarn install) and we kill it after 80s? Doesn't test process end when it's done?

src/extension/src/scripts/run-tests.ts
SahilTara: Oh so we'll never run it locally? If not I wish there was more indication on this via the folder structure.

annielovesjs:
> Did you get this to work with Flask? I'm getting the following with a fresh build of this branch:
> 
> ```
> $ python3 server/server.py || py -3 server/server.py || python server/server.py
> 
> [1]
> Traceback (most recent call last):
> [1]   File "server/server.py", line 1, in <module>
> [1]     from flask import Flask, jsonify, make_response, send_from_directory
> [1] ModuleNotFoundError: No module named 'flask'
> 
> D:\other\WTS\WebTemplateStudio\src\extension\out\scripts\run-tests.js:89
>                 throw new Error("Error found in project");
> ```

I added the command to install python dependencies for the projects with flask. It should work for you now :)

src/extension/src/scripts/run-tests.ts
annielovesjs: Oh I just forgot to remove this haha it's gone now

src/extension/package.json
annielovesjs: I think we should leave it in case we add other tests later. However, I did make this change so you only have to run one command to run both scripts:
    "template-tests": "node ./out/scripts/generate-test && node ./out/scripts/run-tests",
and removed "generate" and "run-tests"

src/extension/src/scripts/run-tests.ts
annielovesjs: yupp! just changed it to const

src/extension/src/scripts/run-tests.ts
annielovesjs: Good point - it's deleted now!

src/extension/src/scripts/run-tests.ts
annielovesjs: You raise a valid point but there's actually a couple of reasons why I didn't follow that approach. 

The generate-test script is for populating template_test folder after it's executed (which is empty in the beginning of every test) and its additional purpose it to make sure that coretsapi also has no problems because otherwise, it would fail to properly generate all the projects with its corresponding pages. 

The template_test folder is emptied after run-tests finishes so we don't have unnecessary files that make the build fail. 

When you run ./build if the generated projects are there, it will fail. So that's also why it is deleted afterwards.

src/extension/src/scripts/run-tests.ts
annielovesjs: it actually doesn't terminate if I do execSync so I have to use a timeout as well

src/extension/src/scripts/run-tests.ts
annielovesjs: wow that actually cleans up my code so much

annielovesjs: Just added it, thanks for the suggestion!

src/extension/src/scripts/run-tests.ts
SahilTara: That's really odd.  Have you tried doing await testProcess, same with await serverProcess? Or is it not worth promisifying?

src/extension/src/scripts/run-tests.ts
annielovesjs: Yeah, it's only going to run on powershell so it'll be fine

src/extension/src/scripts/run-tests.ts
annielovesjs: Hmm ok, I can create it in another pr

streetcoder123 (commented):
Suggestion: Do you think it might make more sense to store the generated projects in the OS default directory for temporary files. `os.tmpdir` might help you with this.

SahilTara (approved):

piotrmark (approved):

src/extension/src/scripts/run-tests.ts
annielovesjs: Yeah, I tried to do a promise before but since this particular process doesn't terminate by itself so I have to use a timeout to force termination after it is finished

jasmingoh (approved):

Issue #810, opened 2019-07-02
Flask Feedback


streetcoder123:
This issue has all the current issues with Flask application that are required to be dealt with before it is released out of preview mode.

- [x] All the import statements should be in specific order in the generated project
- [x] Remove unused import statements from the templates
- [x] Replace all instances of import * with actual imports that you want to have
- [x] Formatting of the return statements should not be on multiple lines
- [x] Some of the methods could be refactored to be more readable
- [x] Convert all the comments to complete sentences
- [x] The conditionals should not have parenthesis wrapped around them
- [x] Change `destroy` function in sql/mongo to `delete` for consistency

PR #835, opened 2019-07-05
Change 'destroy' function in mongo/sql files to delete

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/30478

**Changes**
- Change destroy function in sql/mongo to delete for consistency

GitHub Issue #810


SahilTara (approved):

jasmingoh:
Could you clarify the consistency you mentioned? It seems the node templates use `destroy`

streetcoder123:
@jasmingoh Deleting items is mostly done using the delete command (I know this for Mongo for sure). Christopher mentioned that we should be consistent with that so that the user is not confused. In fact, we should change this in Node as well. We can do that later though.

jasmingoh:
@njindal239 sounds good, please create a backlog item so we won't forget

Tanya0609 (approved):

jasmingoh (approved):

Issue #718, opened 2019-06-12
Place character limit on Project Name


xtinah-w:
Bug bash revealed that you can put in too many characters in the project name which breaks the template generation process.

PR #836, opened 2019-07-05
Limit project name characters

jasmingoh:
This PR addresses #718 ([ADO task](https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/30494))

Users can only enter a project name up to 50 characters in the input field. If their project name is 50 characters long, an error message will show and they will not be able to proceed to the next page:

This is also address the concern that @SahilTara brought up in https://github.com/microsoft/WebTemplateStudio/pull/824#discussion_r300796350


src/client/src/utils/validateName.ts
```diff
@@ -1,5 +1,7 @@
 import { defineMessages, FormattedMessage } from "react-intl";
 
+const PROJECT_NAME_CHARACTER_LIMIT = 50;
```
SahilTara: just curious, why 50?

jasmingoh: Our thought is that developers wouldn't put in such a long project name, and Visual Studio also limit the project name to 50 characters.

xtinah-w: We also based the project name input off of the Windows Template Studio char limit for project name!

SahilTara: Very small nit that comes from this though, can we add this to the constants file, and use the constant here and ProjectNameAndOutput?

SahilTara (approved):

jsondoo (approved):

xtinah-w (approved):
Looks good and it works on my system

PR #839, opened 2019-07-08
Update Code Owners List for PR approval

markAtMicrosoft:
QUick change to add @annielovesjs as a PR approver for this project.  Also removed Trevor and Clint


.github/CODEOWNERS
```diff
@@ -1,2 +1,3 @@
 # At least one of the code owners below will be required on each PR:
-* @markAtMicrosoft @sibille @mvegaca @dgomezc @smmatte @trevorNgo @jcoc611-microsoft @piotrmark @MehaKaushik
+* @markAtMicrosoft @sibille @mvegaca @dgomezc @smmatte @trevorNgo @jcoc611-microsoft @piotrmark @MehaKaushik @annielovesjs
```
jasmingoh, line 2: It seems like Trevor is still in the list.. also, Meha is no longer our coach, should her name still be in the list?

jasmingoh (changes requested):

.github/CODEOWNERS
markAtMicrosoft: The list got updated by somebody else and yes, trevor is still in there.  I'm ok with it for now.  The big thing was to get Annie on boarded.

markAtMicrosoft: I'll confirm the names before I complete this tomorrow.  Might update then but not tonight.

SahilTara: Yeah I think this is due to us updating the code owners a while back,  #781

PR #841, opened 2019-07-08
Right Sidebar - Licenses and Feedback Modals

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/30115

**Changes**
- Remove plain licenses and feedback links from right side bar.
- Add modals for licenses and feedback into the right side bar.

**Expected Right Side Bar**
image


src/client/src/containers/About/index.tsx
```diff
@@ -46,17 +46,6 @@ const About = ({ versions, intl }: Props) => {
           Web Template Studio
         </a>
       </p>
-      <p>
-        <a
-          className={styles.link}
-          href="https://github.com/Microsoft/WebTemplateStudio/issues"
-        >
-          <FormattedMessage
```
piotrmark, line 54: Feels like we can remove FormattedMessage from imports as well :)

piotrmark (dismissed):

src/client/src/containers/About/index.tsx
streetcoder123: Agreed. There were a couple of other unused imports as well. Refactored it. :)

piotrmark (approved):

jasmingoh:
Just talked to @njindal239 about the "View License" button on the home page when the user first launch the extension, since the modal is empty. This will be resolve once #852 is in.

jasmingoh (approved):

annielovesjs (approved):

PR #842, opened 2019-07-08
Azure Functions Form Update Part 2

jsondoo:
**Changes**
- Adds radio button to toggle between 'Choose Existing' and 'Create new resource group for me'
- `chooseExistingRadioButtonSelected` state added to store (we need to save this state in store since user can decide to edit it later)
- CSS tweaks per Christina's suggestions
- If 'Create new resource group' is selected, it will send an empty string to the extension for resource group name, which will handle automating of resource groups as per #824. Resource group will be created in `Central US` and Functions should be deployed in selected location.

formpart2


jasmingoh:
**Note for other reviewers**: Please avoid deploying Azure Functions to locations where there is no pricing information (here's the link: https://azure.microsoft.com/en-us/pricing/details/functions/).
E.g. deploying Azure Functions to "Canada East" will fail, because it is not supported there. 

Here's the GitHub issue for the issue: #847

src/client/src/containers/AzureFunctionsModal/messages.ts
```diff
@@ -93,5 +93,17 @@ export const messages = defineMessages({
   numFunctionsSubLabel: {
     id: "azureFunctionsModal.numFunctionsSubLabel",
     defaultMessage: "Select the number of endpoints you want for your web app"
+  },
+  chooseExisting: {
+    id: "azureFunctionsModal.chooseExisting",
+    defaultMessage: "Choose existing"
```
jasmingoh, line 99: Can the variable name be more specific? e.g. chooseExistingResourceGroup ~~, with default message "Choose from existing resource groups in the subscription"~~

src/client/src/translations/en.json
```diff
@@ -20,8 +20,11 @@
   "azureFunctionsModal.ariaResourceGroupLabel": "Resource Group Drop Down",
   "azureFunctionsModal.ariaRuntimeStackLabel": "Runtime Stack Drop Down",
   "azureFunctionsModal.ariaSubscriptionLabel": "Subscription Drop Down",
+  "azureFunctionsModal.chooseExisting": "Choose existing",
```
jasmingoh: Be specific here too

src/client/src/translations/en.json
```diff
@@ -20,8 +20,11 @@
...
   "azureFunctionsModal.createFunctionApp": "Create Function Application",
   "azureFunctionsModal.createNew": "Create New",
+  "azureFunctionsModal.createNewResourceGroupForMe": "Create new resource group for me",
```
jasmingoh: can this be createNewResourceGroupDisplayMessage?

src/client/src/translations/en.json
```diff
@@ -20,8 +20,11 @@
...
+  "azureFunctionsModal.createNewResourceGroupForyou": "We will automatically create a resource group for you",
```
jasmingoh: and this to createNewResourceGroupSelectedDisplayMessage

src/client/src/translations/whitelist_en.json
```diff
@@ -44,6 +44,9 @@
   "azureFunctionsModal.appNameSubLabel",
   "azureFunctionsModal.runtimeStackSubLabel",
   "azureFunctionsModal.numFunctionsSubLabel",
+  "azureFunctionsModal.chooseExisting",
+  "azureFunctionsModal.createNewResourceGroupForMe",
+  "azureFunctionsModal.createNewResourceGroupForyou",
```
jasmingoh: Update here too

jasmingoh (changes requested):

src/client/src/containers/AzureFunctionsModal/messages.ts
jsondoo: This is for the radio button though, I thought we agreed on this :(

jsondoo: I will change the variable name though

jasmingoh (approved):

annielovesjs (approved):
Works for me now!

src/client/src/containers/AzureFunctionsModal/index.tsx
```diff
@@ -236,9 +238,25 @@ const AzureFunctionsResourceModal = (props: Props) => {
     }
   }, [azureFunctionsFormData.appName.value]);
 
+  /*
+   * Listens on radio button change to update button status
+   */
+  React.useEffect(() => {
```
SahilTara, line 244: YES HOOKS 😄

src/client/src/containers/AzureFunctionsModal/index.tsx
```diff
@@ -503,7 +605,9 @@ const mapStateToProps = (state: AppState): IStateProps => ({
   appNameAvailability:
     state.selection.services.azureFunctions.appNameAvailability,
   isValidatingName: state.selection.isValidatingName,
-  selection: getFunctionsSelection(state)
+  selection: getFunctionsSelection(state),
+  chooseExistingRadioButtonSelected:
```
SahilTara, line 609: shouldn't be adding this here, instead: https://github.com/microsoft/WebTemplateStudio/blob/dev/src/client/src/selectors/azureFunctionsServiceSelector.ts then use selection object.

jsondoo: I tried this before, but I decided not to use this because `chooseExistingRadioButtonSelected` is a boolean which is not really in the form `label` and `value`, and the dropdown types were really not very flexible. It would require a lot more type changes to get this done, so I could do this in a separate PR if needed. 

Do you happen to know why `selection` is an array by the way? @SahilTara

SahilTara: Think it was because at one point it was able to accept multiple functions apps or something

SahilTara: Imo the array should be killed off if you dont plan to support multiple function apps being deployed at once. (Not in this pr)

SahilTara (approved):

piotrmark (approved):

PR #845, opened 2019-07-08
Create resource group for Azure CosmosDB

jasmingoh:
Related PBI: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/30315

The purpose of this PR is to create resource group for user if they choose to put the Azure Functions and/or CosmosDB they create in a new resource group.

Client side changes will be in another PR (the client will be sending an empty string for `functions.resourceGroup` and/or `cosmos.resourceGroup` field in the payload when user choose a new resource for the respective services). 

So, to test this PR:
- Add **_(1)_** `payload.functions.resourceGroup = "";` and/or ### _(2)_ `payload.cosmos.resourceGroup = "";` before line 92 in `generationExperience.ts`. This will force the method that deploys creation of resource group to be kicked off. 
- Launch the wizard
- Take note of the project name you entered in the first page of the wizard. 
- Add Azure Functions and/or CosmosDB when creating your project
- Note: **if you added _(1)_**, the resource group you select in the Azure Functions form will not matter, because it will be deployed to the new resource group, but please check to see if your CosmosDB is deployed to the selected resource group, same goes for **_(2)_**.
- Note: **if you added _(1)_ and _(2)_**, the resource group you select in the Azure Functions form and CosmosDB form will not matter, because both of them will be deployed to the new resource group.
- Go onto Azure portal and check if a new resource group has been created. The name of the new resource group should start with the project name your entered
- Make sure that the services you chose are deployed to the right resource groups

**Note**: The arm template required for resource group in the generated project will also be addressed in another PR.


src/extension/src/azure/azureServices.ts
```diff
@@ -287,26 +287,58 @@ export class AzureServices extends WizardServant {
     }
   }
 
-  public static async deployResourceGroup(payload: any): Promise<any> {
+  public static async generateDistinceResourceGroupSelections(
```
piotrmark: Typo here? :)

piotrmark (dismissed):

src/extension/src/azure/azureServices.ts
jasmingoh: good catch :)

streetcoder123:
@jasmingoh As discussed with you, check how are you making the subscriptions array. The subscription array should only include the those distinct subscriptions for which you receive an empty payload of resource group.

src/extension/src/azure/azure-resource-group/resourceGroupModule.ts
```diff
@@ -47,29 +48,42 @@ export class ResourceGroupDeploy {
...
-    userSubscriptionItem: SubscriptionItem
+    userSubscriptionItems: SubscriptionItem[]
   ): Promise<string> {
-    this.setAzureResourceClient(userSubscriptionItem);
-
     let generatedName = this.generateResourceGroupName(name);
-    let isValid = await this.validateResourceGroupName(generatedName);
+    let isValid: boolean = true;
+    // this allows the generated name to be validated against multiple subscriptions
+    userSubscriptionItems.forEach(async userSubscriptionItem => {
```
SahilTara: nit why repeat this code here and in lines 79-86? The lines look very similar if not the same.

src/extension/src/azure/azureServices.ts
```diff
@@ -287,26 +287,66 @@ export class AzureServices extends WizardServant {
...
+      );
+      allSubscriptions.push(AzureServices.usersCosmosDBSubscriptionItemCache);
+    }
+    const allDistinctSubscriptions: SubscriptionItem[] = [... new Set(allSubscriptions)];
+    const generatedName: string = await AzureServices.AzureResourceGroupProvider.generateValidResourceGroupName(
+          projectName,
+          allDistinctSubscriptions
+        );
+    let tempResourceGroupSelection: ResourceGroupSelection;
+    allDistinctSubscriptions.forEach(subscription => {
```
SahilTara: Can get rid of result on line 294, line 314 variable and  replace with 
```ts
return allDistinctSubscriptions.map(subscription =>  AzureServices.generateResourceGroupSelection(generatedName, subscription));
```

src/extension/src/azure/azureServices.ts
```diff
@@ -287,26 +287,66 @@ export class AzureServices extends WizardServant {
...
-    const generatedName = await AzureServices.AzureResourceGroupProvider.generateValidResourceGroupName(
-      projectName,
-      this.usersFunctionSubscriptionItemCache
-    );
+    return result;
+  }
 
-    const resourceGroupSelection: ResourceGroupSelection = {
-      subscriptionItem: AzureServices.usersFunctionSubscriptionItemCache,
+  public static functionsSelectedNewResourceGroup(payload: any): boolean {
```
SahilTara, line 321: is the payload really any here? Does it need to be public?

jasmingoh: The payload is what the client sends to the extension. Do you have any suggested type that the payload should have?
Yes, it is used in the generateExperience too, so it has to be public.

src/extension/src/azure/azure-resource-group/resourceGroupModule.ts
jasmingoh: because I need to know if the first generated name is valid, if it is not, then the function will re-generate a name and re-validate the new name within the loop

SahilTara: Why not just go into the loop right away by setting isValid to false?

src/extension/src/azure/azureServices.ts
SahilTara: if its used in generateExperience never mind

SahilTara (dismissed):

jsondoo:
@jasmingoh I tried adding Add (1) payload.functions.resourceGroup = ""; and/or ### (2) payload.cosmos.resourceGroup = ""; and then deploying both cosmos and functions. It deploys both (checking on Azure portal) but client side it gives me an error for CosmosDB
image

piotrmark (dismissed):

jasmingoh:
@jsondoo my bad, it should work now, there was a typo :(

jsondoo (approved):
cosmos works now!

SahilTara (approved):

annielovesjs (approved):

Tanya0609 (approved):

PR #859, opened 2019-07-09
Frameworks Mock Data for mockVSCodeapi

jsondoo:
**Changes**
Before we used to call CoreTS directly, but since now we call the extension, we need to mock the payload the extension sends back in the mockVSCodeapi

**How To Test**
Check that frameworks page renders the framework cards in browser


streetcoder123 (commented):
Works :)

streetcoder123 (approved):

annielovesjs (approved):
all good now

Tanya0609 (approved):

PR #862, opened 2019-07-09
CosmosDB Form Update Part 2

### LINKED ISSUE #455: Creating a new Azure resource group does not automatically populate the modal with the new group
1. Press create new resource group
2. Create a new resource group
3. You now need to leave the modal and return to see your new resource group

Suggestion: Have a "Create New" option within the dropdown

jsondoo:
closes #455 

**Changes**
- Adds radio button to toggle between 'Choose Existing' and 'Create new resource group for me'
- `chooseExistingRadioButtonSelected` state added to store (we need to save this state in store since user can decide to edit it later)
- If 'Create new resource group' is selected, it will send an empty string to the extension for resource group name, which will handle automating of resource groups as per #824. Resource group will be created in `Central US` and CosmosDB should be deployed in selected location

cosmospart2

should be tested and merged after #845


SahilTara (commented):
Looks good for the most part, only concerning bit is that there seems to be a lot of duplicate logic between functions and cosmos, do you think this will be addressed in a different PR (since it feels out of scope for this PRs goals). Other than that i'm happy with this PR.

jsondoo:
@SahilTara Yea, I agree there's a lot of duplicate logic. We still have one more form to work on so I could do some clean up afterwards since I don't wanna spend too much time on refactoring right now.

SahilTara:
This should also close #455 right?

piotrmark (approved):

SahilTara (approved):

jasmingoh (approved):
Please only merge after #845 is merged, else the "Create a new resource group for me" won't work

PR #864, opened 2019-07-09
Add Remove Page Functionality using '-' symbol

### LINKED ISSUE #863: Add a '-' button to remove pages from the pages page
Currently, the only way to remove pages is using the right side bar, which is not very user friendly. So, we need to have both '+' and '-' buttons in the page cards to allow users to add and remove pages.

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/30713

**Changes**
- Add a svg for '-' button into the page cards
- Add ability to decrement the number of pages using the button
- Not allow to delete all the pages (At least 1 page should be generated)

**Expected Look**
image

closes #863


streetcoder123:
There is one issue with this one that I have not been able to figure out. 
image

For the text that shows number of pages, if the number of pages are 0, the text is little smaller than if it is non-zero. I tried to play around with it but to no success. Does someone has an insight on it.

I am suspecting that it might be related to the fact that `<div> 0 </div>` is replaced as `<dviv> 0 </div> ==$0`. I am not sure though what it means.

SahilTara:
image
Visual bug light mode.

Also @njindal239 the font difference has to do with styles.selected, font size is set to 15px for selected. Look at the css file!

SahilTara:
Also when using keyboard only navigation theres no way of telling you have selected + or - button currently

SahilTara:
Also another piece of feedback the "-" button bounces around a lot as you increase the numbers. Which isn't ideal.

piotrmark (commented):
Should we hide '-' button when there's 0 pages selected?

src/client/src/components/SelectableCard/index.tsx
```diff
@@ -114,14 +118,24 @@ const SelectableCard = ({
           />
         </Link>
         <div className={styles.pageButtons}>
+          {isPagesSelection && (
+            <button
+              className={classNames(styles.cardCount, styles.countButton)}
+              onClick={() => {
+                removePage(cardNumber);
```
ngkelly700: nit: I think this can be on one line like:
`() => removePage(cardNumber)`

ngkelly3: nit: I think you can make this into one line, i.e.
`() => removePage(cardNumber)`

src/client/src/assets/subtract.svg
```diff
@@ -0,0 +1,4 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="12" height="3" viewBox="0 0 9 2" fill="none">
```
jcoc611-microsoft: Just curious, what was the thought process on using SVGs vs just using the characters `+` and `-`?

src/client/src/containers/SelectOption/index.tsx
```diff
@@ -219,6 +240,24 @@ class SelectOption extends React.Component<Props, ISelectOptionState> {
...
 
+  public removePage(cardNumber: number) {
+    const {
+      options,
+      currentCardData,
+      cardTypeCount,
+      handleCountUpdate
+    } = this.props;
+    const { internalName } = options[cardNumber];
+    if (
```
jcoc611-microsoft, line 253: I know it's mostly for sanity, but is it necessary to perform these checks both here and in the `removeOption` function?

src/client/src/assets/subtract.svg
streetcoder123: Firstly, I used the + and - buttons rather than SVGs. However, the designs and styling resembled much more closely to the actual designs by the designer using these SVGs. 
Though, now I am facing a problem on how to change the color of SVGs based on the VSCode color theme :p

src/client/src/containers/SelectOption/index.tsx
streetcoder123: Ideally it is not at all necessary. However, it just allows to make more independent functions that can be reused easily.

streetcoder123:
@SahilTara Thanks for the feedback. I have addressed your feedback on keyboard usage and button bounces. I am still looking into changing SVG colors for light color theme.

src/client/src/containers/SelectOption/index.tsx
```diff
@@ -144,6 +144,27 @@ class SelectOption extends React.Component<Props, ISelectOptionState> {
...
+    cardNumber: number,
+    cardCount: number,
+    internalName: string
+  ) {
+    const { selectedCardIndices, currentCardData, selectOptions } = this.props;
+    if (selectOptions && currentCardData && currentCardData.length > 1) {
+      const size = currentCardData.length;
+      for (let i = size - 1; i >= 0; i--) {
+        if (currentCardData[i].internalName === internalName) {
+          currentCardData.splice(i, 1);
```
jcoc611-microsoft: I think `splice` might not be good here as the `currentCardData` is part of the state (`this.props`) and should be immutable/constant.

streetcoder123: If it is manipulating the state object, it is definitely needs to be corrected. However, please guide me if I am wrong. In select pages, I guess, we already destructure the state object before returning them as props. So, I think we do not need to create a copy of element here. 
Am I thinking right?

streetcoder123:
> Should we hide '-' button when there's 0 pages selected?

@piotrmark The UI does not look good if we do that. This is more intuitive (I think) as we have a cart system. This was a feedback when I implemented just the '+' button that it looks odd (since there was no '-' button then).

src/client/src/components/SelectableCard/index.tsx
```diff
@@ -114,14 +118,22 @@ const SelectableCard = ({
           />
         </Link>
         <div className={styles.pageButtons}>
+          {isPagesSelection && (
+            <button
+              className={classNames(styles.cardCount, styles.countButton)}
+              onClick={() => removePage(cardNumber)}
+            >
+              {subtract && <Subtract className={styles.subtractSVG} />}
```
jsondoo, line 126: Looks good! 
We should have an `alt` tag for this svg

src/client/src/assets/subtract.svg
jcoc611-microsoft: Check this one out: https://css-tricks.com/cascading-svg-fill-color/

src/client/src/containers/SelectOption/index.tsx
jcoc611-microsoft: This eventually creates a new object state, but it is still best practice not to touch the current state. It should be a copy of the array.

annielovesjs (approved):
looks good in both light mode and dark mode now :D

jsondoo (approved):

SahilTara (approved):

PR #865, opened 2019-07-10
Clean up client side so that unused functions are not calling CoreTS API anymore

jasmingoh:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/30717

For getFeature:
- Found out that it is not being used in the project, so it is removed

For getProjectType:
- We removed the Project Type page in #741 so there is no use for the method anymore

### DIFF
--- a/src/client/src/actions/wizardContentActions/getProjectTypes.ts
+++ b/src/client/src/actions/wizardContentActions/getProjectTypes.ts
@@ -1,10 +1,6 @@
-import EngineAPIService from "../../services/EngineAPIService";
 import { IMetadata } from "../../types/metadata";
 import { IOption } from "../../types/option";
-import { getProjectTypesSuccess } from "./getProjectTypesSuccess";
 import getSvgUrl from "../../utils/getSvgUrl";
-import WizardContentActionType from "./wizardContentActionType";
-import { Dispatch } from "react";
 
 function getMetadataFromJson(items: any[]): IMetadata[] {
   return items.map<IMetadata>(val => ({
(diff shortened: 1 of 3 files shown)


src/client/src/actions/wizardContentActions/getProjectTypes.ts
```diff
@@ -6,29 +6,6 @@ import getSvgUrl from "../../utils/getSvgUrl";
 import WizardContentActionType from "./wizardContentActionType";
 import { Dispatch } from "react";
 
-// thunk
```
SahilTara, line 9: Do you need this action at all then? (File)

piotrmark (dismissed):
Removing code <3

src/client/src/actions/wizardContentActions/getProjectTypes.ts
jasmingoh: @SahilTara you're right. 
Honestly, I'm not sure if this file is needed- The methods are not used at all, but the reason I did not remove it is because we don't know if there will be other project types added in the future and

jcoc611-microsoft: @Tanya0609 did you implement getting project types from the extension? If so you should update this file. We should keep the action in case there are more project types in the future as Jasmin mentioned.

src/client/src/services/EngineAPIService.tsx
```diff
@@ -12,41 +12,6 @@ export default class EngineAPIService {
     }
   }
```
jcoc611-microsoft, line 13: This entire file should be removed. If we haven't implemented a way to get the pages from the extension we should do that first.

jasmingoh:
Closing this PR for now, will open again after #874 is merged

jasmingoh:
Reopened as a new PR here: #891

PR #867, opened 2019-07-10
Cleaned up some code giving warnings

SahilTara:
This PR Addresses the following issues:
  - Some of the code was giving lots of "unused" warnings and such.
  - TopNavBar code was highlighted almost completely yellow. 
  - Weird glitch happening where a clicked disabled nav item would be highlighted blue for a split second.


jasmingoh:
just curious, I still see that there are "unused" warnings, how did you decide which ones to fix in this PR?

SahilTara:
> just curious, I still see that there are "unused" warnings, how did you decide which ones to fix in this PR?

I forgot some, I don't want to touch SelectOptions since @njindal239 is touching that file a decent amount.

streetcoder123:
Thanks @SahilTara 
Moving code sometimes introduces a lot of unused warnings. I think this is another area we need to be careful about since it does not come up as error.
I shall also be looking into some of these as I touch files :)

streetcoder123 (approved):

annielovesjs (approved):
Thanks for the clean up :)

jsondoo (approved):

PR #868, opened 2019-07-10
Added keyshortcut

### LINKED ISSUE #472: Create a key shortcut to run the extension
Create a shortcut, example: `Ctrl+Shift+A` mapped to `Web Template Studio : Launch`

SahilTara:
This PR allows the extension to be launched via ctrl+shift+a or cmd+shift+a.

This was the key combination suggested in the issue. We can change it to whatever you want though. 

closes #472


jcoc611-microsoft:
I can't imagine users needing to launch the wizard very often. Is there enough justification to pollute the user's key bindings?

SahilTara:
Its an easier to access method, and it was requested by one of our very early users. Hence the incomplete issue. Most extensions (ive seen) use some type of key binding for their main command. @jcoc611-microsoft What do you think should be done to validate or invalidate the addition of a shortcut?

trevorNgo:
I was really looking into finding a way to define a **settings field** for keyboard shortcuts in the _contributes.configuration.properties_... no dice

src/extension/package.json
```diff
@@ -49,6 +49,13 @@
         "category": "%webTemplateStudioExtension.commands.wts%"
       }
     ],
+    "keybindings": [
+      {
+        "command": "webTemplateStudioExtension.wizardLaunch",
+        "key": "ctrl+shift+a",
```
trevorNgo, line 55: I would consider leaving this blank so that it would appear in the keyboard shortcuts view but up to individual users to define: 
image

SahilTara: So you think its better to just give the user the ability but define no default? If so the extension already provides the ability. This PR just defines a default for the user. 

If you think its better to just leave the thing unbound then i'll close this PR and mark the issue as wont fix or something.

jsondoo (approved):

jsondoo:
Is it possible to have this just for local development? I agree that we shouldn't pollute other users bindings but this seems useful for us

SahilTara:
> Is it possible to have this just for local development? I agree that we shouldn't pollute other users bindings but this seems useful for us

Hmm yeah you could do it manually if needed do you suggest I close this PR + close the related issue and mark as wontfix? @jsondoo

jsondoo:
@SahilTara yea, we discussed this with @njindal239 and @jcoc611-microsoft and we think this would be nice to have for just us. I think we can close this PR

PR #869, opened 2019-07-10
remove event keycode

### LINKED ISSUE #817: keyCode API is deprecated

annielovesjs:
closes #817

Changes:
- replace event.keyCode with event.key for better readibility

How to test:
- go on extension and make sure the corresponding keys still trigger the same events


src/client/src/translations/en.json
```diff
@@ -2,7 +2,6 @@
   "FunctionNameError.invalidFunctionStartLetter": "Name may only start with letters or numbers",
   "FunctionNameError.invalidRegex": "Name may only contain letters, numbers or dashes",
   "about.about": "About",
-  "about.reportAnIssue": "Give Feedback or Report an issue",
```
jcoc611-microsoft: Why was this one removed?

src/client/src/components/Details/index.tsx
```diff
@@ -51,7 +51,7 @@ const Details = ({
     </React.Fragment>
   );
   const keyDownHandler = (event: any) => {
```
SahilTara: change all any to React.KeyboardEvent happens in multiple places

SahilTara:
Also should add closes #817

jasmingoh:
@annielovesjs make sure to run `npm manage:translations` !

src/client/src/translations/en.json
streetcoder123: I think this needs to be removed. When we changed the licenses from a plain list to modals, this was changed to `"about.reportAnIssueLabel": "Give Feedback"`, which is already present in the file.
I am not very sure though.
@annielovesjs Do you have any comments on this one?

src/client/src/components/Details/index.tsx
annielovesjs: done :)

annielovesjs:
> @annielovesjs make sure to run `npm manage:translations` !

haha I did xD

src/client/src/translations/en.json
annielovesjs: I don't think manage translations works properly so I have to manually change the text in translation files. I just added this one back

SahilTara (dismissed):

src/client/src/components/Details/index.tsx
```diff
@@ -50,8 +50,8 @@ const Details = ({
       <br />
     </React.Fragment>
   );
-  const keyDownHandler = (event: any) => {
-    if (event.keyCode === 13 || event.keyCode === 32) {
+  const keyDownHandler = (event: React.KeyboardEvent<HTMLDivElement>) => {
+    if (event.key === "Enter" || event.key === " ") {
```
streetcoder123: Since "Enter" is being used at multiple places, do you think it would make more sense to make it a constant?

SahilTara: Actually, yes that is a good idea, it makes things more safe (Less prone to typos).
Add something like this in the constants file:
```ts
export enum Key {
  Enter = "Enter",
  Space = " ",
  Tab = "Tab"
}
```

piotrmark (dismissed):
+1 to Sahil's comment about consts. 
Other than that, looks good :)

src/client/src/translations/en.json
annielovesjs: manage translations didn't effect these files for some reason so I just added it back in

src/client/src/components/Details/index.tsx
annielovesjs: good point - just added :)

jcoc611-microsoft (dismissed):
Nice!

piotrmark (dismissed):

src/client/src/utils/constants.ts
```diff
@@ -49,7 +49,13 @@ const COSMOS_APIS = {
...
 enum FRAMEWORK_TYPE {
   FRONTEND = "frontend",
-  BACKEND = "backend",
+  BACKEND = "backend"
+}
+
+enum KEY_EVENTS {
+  Enter = "Enter",
+  Space = " ",
+  Tab = "Tab"
```
streetcoder123: Can you capitalize these for consistency?

streetcoder123 (approved):

piotrmark (approved):

jasmingoh (approved):

PR #873, opened 2019-07-10
azure page redesign

annielovesjs:
Changes:
- Added an optional step flag/button that leads you to the last page (improves visibility)
- styling changes for azure services page
- adjusted padding for every page 

How to test:
- make sure the button always routes to the right page 
- should look like this when not logged in: 
notloggedin
- should look like this when logged in:
loggedin


src/client/src/components/LoginCard/index.tsx
```diff
@@ -29,32 +30,24 @@ const LoginCard = ({
...
-          <FormattedMessage id="loginCard.details" defaultMessage="Details" />
-        </Link>
-        <div className={styles.buttonContainer}>
-          <a
-            className={styles.createAccountButton}
-            href="https://azure.microsoft.com/free/"
+    <div className={styles.loginPrompt}>
+      <div className={styles.loginContainer}>
+        <div className={styles.cardTitle}>
+          <img className={styles.icon} src={svgUrl} alt="" />
```
jcoc611-microsoft, line 36: Out of curiosity, why are we using `svg` url instead of directly importing the `.svg` file like in Nipun's `-`/`+` PRs?

jsondoo: The way we do svgs in the client right now is really inconsistent. Sometimes we use `src` and sometimes import it as a component. So probably a good idea to clean this up sometime

jsondoo:
@annielovesjs should the text for the banner change to blue on hover? happens on extension but not on browser

src/client/src/components/LoginCard/index.tsx
SahilTara: From what I can remember I think we did svgUrl for things that are/can be dynamic, and raw import for static assets. Example: page icons for dynamic. Example: checkmark for static. So that brings up the question does the icon here change dynamically? Or will it always be the same icon?

annielovesjs: As of now, it will always be the same icon because we only use it once. But it would be helpful if we need to reuse the same component but change the svg

src/client/src/mockData/azureServiceOptions.ts
```diff
@@ -55,29 +64,31 @@ export const messages = defineMessages({
...
+    title: azureMessages.azureFunctionsTitle,
     internalName: WIZARD_CONTENT_INTERNAL_NAMES.AZURE_FUNCTIONS,
-    longDescription: messages.azureFunctionsLongDescription,
-    body: messages.azureFunctionsCardBody
+    longDescription: azureMessages.azureFunctionsLongDescription,
+    body: azureMessages.azureFunctionsCardBody
   },
   {
     author: "Microsoft",
+    type: "Cloud Database",
```
SahilTara: Perhaps make the types into constants/into an enum in case we ever add more db types/hosting types?

piotrmark (dismissed):
Looks good :)

src/client/src/containers/AzureSubscriptions/index.tsx
```diff
@@ -110,29 +118,69 @@ class AzureSubscriptions extends React.Component<Props, IState> {
...
+    title: any
+  ) {
+    const { formatMessage } = this.props.intl;
+    return (
+      <div
+        className={classnames(styles.servicesContainer, {
+          [styles.overlay]: !isLoggedIn
+        })}
+      >
+        {azureServiceOptions.map(option => {
```
SahilTara: I think this is setup a bit wrong.

Notice this is what happens if I change the type of both services to the same thing:
image

This would mean if you add an additional service of a type already provided you will encounter the same bug (For example: App service as a cloud hosting platform).

Instead I think your code should be something like the following
```ts
<div className={styles.servicesCategory}>
  <div className={styles.categoryTitle}>{type}</div>
  <div className={styles.categoryDescriptor}>{formatMessage(title)}</div>
  {azureServiceOptions.map(option => {
    if (option.type === type) {
      return (
        <div
          key={JSON.stringify(option.title)}
          className={classnames(styles.subscriptionCardContainer)}
        >
          <Card
            option={option}
            buttonText={this.addOrEditResourceText(option.internalName)}
            handleButtonClick={this.getServicesModalOpener(option.internalName)}
            disabled={!isLoggedIn}
            handleDetailsClick={setDetailPage}
            useNormalButtons={this.isSelectionCreated(option.internalName)}
          />
        </div>
      );
    }
  })}
</div>
```

Which is slightly better but the css is a bit messed up, as the cards should be beside each other but they're on a new line here. They should behave like the pages kind of. Where if the screen is too small it will move them to the next row, otherwise they will try to populate the same row. 

image


Now you may notice that if you have an empty type, (a type defined with no elements the title will show up. If you want this to happen leave the above code as is. If not, you can do a filter, then a length check before rendering.

annielovesjs:
> @annielovesjs should the text for the banner change to blue on hover? happens on extension but not on browser

it is fixed now :)

src/client/src/appStyles.module.css
```diff
@@ -13,15 +13,35 @@
...
 .centerViewMaxHeight {
   height: calc(100vh - 51px);
   margin-bottom: 0px;
 }
 
+.optionalFlag {
+  position: absolute;
+  width: 100%;
+  background-color: #0067b8;
+  color: white;
```
jsondoo: Can we use vscode theme variables here? I think white is editor-foreground

src/client/src/appStyles.module.css
```diff
@@ -13,15 +13,39 @@
...
+  text-align: center;
+  text-decoration: none;
+  font-weight: lighter;
+  font-size: 1.4em;
+  padding: 10px 0px;
+}
+
+.optionalFlag:hover {
+  color: white;
```
jsondoo: also here

src/client/src/containers/AzureSubscriptions/index.tsx
annielovesjs: Thanks for bringing this to my attention! I definitely should have tested/accounted for those cases. I made the changes you suggested and to account for only showing the categories in use, I stored the unique option types so I can only render unique categories that have child elements

src/client/src/appStyles.module.css
annielovesjs: hello jason

annielovesjs: I actually didn't use a variable because I needed the colours to be the same in both light and dark mode, but you're right it's generally better practice :)  I found a variable that stays white in all mode so I'm gonna use that instead

src/client/src/containers/AzureSubscriptions/index.tsx
SahilTara: image
Looks better!

src/client/src/mockData/azureServiceOptions.ts
annielovesjs: Sure! Just added it :)

src/client/src/containers/AzureSubscriptions/index.tsx
```diff
@@ -110,29 +118,74 @@ class AzureSubscriptions extends React.Component<Props, IState> {
...
-              handleButtonClick={this.getServicesModalOpener(
-                option.internalName
-              )}
-              disabled={!isLoggedIn}
-              handleDetailsClick={setDetailPage}
-              useNormalButtons={this.isSelectionCreated(option.internalName)}
-            />
-          </div>
-        ))}
+        {serviceTypes.map(option => {
```
SahilTara: may wanna rename this for readability (option) possibly to serviceType?

src/client/src/containers/AzureSubscriptions/index.tsx
```diff
@@ -110,29 +118,74 @@ class AzureSubscriptions extends React.Component<Props, IState> {
...
+            })}
+          </div>
+        </div>
+      </div>
+    );
+  }
   public render() {
     const { isLoggedIn, setDetailPage } = this.props;
+    const serviceTypes: any[] = [];
+    azureServiceOptions.map(option => {
```
SahilTara: think you can just do a .forEach here since you aren't mapping the elements to a new format.

SahilTara (commented):
Looks pretty good after my comments are addressed I will approve.

SahilTara:
Just thought this would help you grow your knowledge base of javascript/typescript, here's another way of going about things:

image

You can also make it more readable by splitting it up:
image

Not saying you should do it like this, its just another way of doing it which I found kind of interesting.

src/client/src/containers/AzureSubscriptions/index.tsx
annielovesjs: I think it makes more sense for it to be plural just because it contains multiple types

SahilTara (dismissed):

annielovesjs:
> Just thought this would help you grow your knowledge base of javascript/typescript, here's another way of going about things:
> 
> image
> 
> You can also make it more readable by splitting it up:
> image
> 
> Not saying you should do it like this, its just another way of doing it which I found kind of interesting.

hey, thanks for the suggestion!  I just made the changes and it looks cleaner/more intuitive now :D

SahilTara:
https://github.com/microsoft/WebTemplateStudio/blob/c0b05ad2d81c895f4301a9b23c0b995f09ed3ad9/src/client/src/containers/CosmosResourceModal/index.tsx#L15

Your code is failing build because of this line you'll need to change it.

annielovesjs:
> https://github.com/microsoft/WebTemplateStudio/blob/c0b05ad2d81c895f4301a9b23c0b995f09ed3ad9/src/client/src/containers/CosmosResourceModal/index.tsx#L15
> 
> Your code is failing build because of this line you'll need to change it.

Yeah, it's a recent merge that just conflicted with mine - should be resolved now

piotrmark (dismissed):

annielovesjs:
>  import { messages as azureModalMessages } from "../../mockData/azureServiceOptions";

Oh turns out it's a recent merge that just happened to conflict with mine, I just resolve the issue

src/client/src/components/TopNavBarLink/index.tsx
```diff
@@ -33,6 +33,7 @@ const TopNavBarLink = ({
       className={styles.container}
     >
       <div
+        aria-label="heyyy"
```
SahilTara: Should this be "heyyy"

annielovesjs: oh shoot sorry I was testing something

annielovesjs: did it on the wrong branch

SahilTara (dismissed):

src/client/src/containers/AzureSubscriptions/index.tsx
```diff
@@ -110,29 +118,72 @@ class AzureSubscriptions extends React.Component<Props, IState> {
...
-              useNormalButtons={this.isSelectionCreated(option.internalName)}
-            />
-          </div>
-        ))}
+        {uniqueServiceTypes.map(serviceType => {
+          return this.getServicesOrganizer(
+            serviceType,
+            isLoggedIn,
+            setDetailPage,
+            messages.hostingTitle
```
streetcoder123: Why do you have messages.hostingTitle for all the Azure services?
`Azure Functions` should have `hostingTitle` but `CosmosDB` should have 'storageTitle`

image

src/client/src/components/TopNavBarLink/index.tsx
jsondoo: heyyy

src/client/src/selectors/postGenerationSelector.ts
```diff
@@ -81,13 +81,13 @@ const servicesToDeploy = (
 ): IAzureServiceStatus => {
   return {
     cosmosdb: {
-      title: messages.cosmosTitle,
+      title: azureMessages.cosmosTitle,
```
streetcoder123, line 84: I am just a little confused here. The azure messages are coming from the mockData. Are these always supposed to come from mockData (I mean even when we are running in extension)? 
If yes, I think we should store this data somewhere else rather than mockData since it is misleading (the data is not actually mock in that case, maybe staticData)

jsondoo: The client gets all the data for azure services page from the mockData (maybe it was supposed to be moved to be retrieved from extension later?). I agree that we should move it or rename it in a different PR maybe

src/client/src/containers/AzureSubscriptions/index.tsx
SahilTara: To fix this i would do something like this.
```ts
const serviceTypeTitle = {
  servicesEnum.DATABASE: message.storageTitle,
  servicesEnum.HOSTING: message.hostingTitle
}
```
Then pass it via:
```ts
serviceTypeTitle[serviceType]
```

(Syntax may be a bit off but that should give you an idea)

annielovesjs: My bad! I just changed it  now so it accounts for both cases with conditionals

src/client/src/selectors/postGenerationSelector.ts
annielovesjs: Yup - will do it in another pr

SahilTara (dismissed):

src/client/src/containers/AzureSubscriptions/index.tsx
```diff
@@ -110,29 +119,81 @@ class AzureSubscriptions extends React.Component<Props, IState> {
...
+            })}
+          </div>
+        </div>
+      </div>
+    );
+  }
   public render() {
     const { isLoggedIn, setDetailPage } = this.props;
+    const serviceTypes = azureServiceOptions.map(option => option.type);
+    const uniqueServiceTypes = serviceTypes.filter(
```
jcoc611-microsoft: NIT
```
const uniqueServiceTypes = [...new Set(serviceTypes)];
```

src/client/src/mockData/azureServiceOptions.ts
```diff
@@ -110,32 +119,39 @@ export const messages = defineMessages({
   }
 });
 
+export const servicesEnum = {
+  HOSTING: "Cloud Hosting",
```
jcoc611-microsoft, line 123: I think ideally this enum would be so that for each item in `serviceEnum` the message `<enum item>Service.title` exists. See the other comment where you actually use the message in AzureSubscriptions container for more context

src/client/src/containers/AzureSubscriptions/index.tsx
```diff
@@ -110,29 +119,81 @@ class AzureSubscriptions extends React.Component<Props, IState> {
...
-              )}
-              disabled={!isLoggedIn}
-              handleDetailsClick={setDetailPage}
-              useNormalButtons={this.isSelectionCreated(option.internalName)}
-            />
-          </div>
-        ))}
+        {uniqueServiceTypes.map(serviceType => {
+          let categoryTitle;
+          switch (serviceType) {
```
jcoc611-microsoft, line 179: If you change the `servicesEnum` you could simply do here `messages[serviceType + 'Title']`

src/client/src/containers/AzureSubscriptions/index.tsx
SahilTara: oh yeah forgot about that way when I gave the split up things example.

src/client/src/mockData/azureServiceOptions.ts
annielovesjs: Do you mean something like this?
messages = {"Cloud Hosting Title": {id: whateverID, defaultMessage: "Host your web app in the cloud", [..]} }

or this: 
const servicesEnum = {
  HOSTING: {name: "Cloud Hosting", message: "Host your web app in the cloud"}
  [...]
}

src/client/src/containers/AzureSubscriptions/index.tsx
annielovesjs: righttt I could have converted it to a set

annielovesjs: just added :)

src/client/tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "typeRoots": ["./node_modules/@types", "./images.d.ts"],
     "target": "es5",
+    "downlevelIteration": true,
```
jsondoo, line 5: What is this for?

annielovesjs: allows me to use the spread operator when I was converting a set to an array

src/client/src/containers/AzureSubscriptions/index.tsx
annielovesjs: will dismiss this for now as approaching it this way does not seem to work with our application

jcoc611-microsoft (approved):

jsondoo (approved):
looks good

jasmingoh (approved):

PR #875, opened 2019-07-10
Forms - Let users know where resource group is created

jsondoo:
@markAtMicrosoft told us that we need to let our users know that their resource group is being created in central US

**Changes**
- string and CSS tweaks
- move duplicate keys

image


SahilTara (dismissed):

streetcoder123 (dismissed):

annielovesjs (dismissed):

src/client/src/translations/en.json
```diff
@@ -58,6 +57,9 @@
   "cosmosDb.longDescription": "Azure Cosmos DB is Microsoft’s proprietary globally-distributed, multi-model database service for managing data on a global scale. It offers a variety of APIs for your database including Azure Table, Core (SQL), MongoDB and Gremlin (GraphQL). Web Template Studio offers you the functionality to deploy a Cosmos DB instance from the wizard itself and select an initial location to deploy your database with the ability to scale it to multiple locations at a future time. As an added feature, deploying with the MongoDB API enables you to quickly connect the project Web Template Studio generates to your database instance.",
   "cosmosDb.originalTitle": "CosmosDB",
   "cosmosDb.title": "Cosmos DB",
+  "cosmosModal.chooseExistingResourceGroup": "Choose existing",
+  "cosmosModal.createNewResourceGroupDisplayMessage": "Create new resource group for me",
```
jcoc611-microsoft: Should these messages be the same for cosmos modal and functions modal? Like use the same key and reuse across modals?

jcoc611-microsoft (changes requested):

src/client/src/translations/en.json
jsondoo: i moved the keys that are duplicated into one place, since we will probably be reusing them for App Services form

jcoc611-microsoft (approved):
Nice!

jasmingoh (approved):

piotrmark (approved):

PR #877, opened 2019-07-10
Next Arrow is not Accessible in VSCode Light Mode

### LINKED ISSUE #876: Next Arrow is not Accessible in VSCode Light Mode
The svg arrow added to the 'Next' button in the footer is not accessible in VSCode light theme.

**In Dark theme:**
image

**In Light Theme:**
image

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/30775

**Change**
- Add style for Next Arrow in the Footer to be accessible in VSCode Light themes

closes #876


src/client/src/containers/Footer/styles.module.css
```diff
@@ -41,6 +41,10 @@
   color: #ffffff;
 }
 
+.nextIcon path {
+  fill: var(--vscode-editor-foreground);
```
jcoc611-microsoft, line 45: Would it break anything if we simply had something like this in a global css file? To prevent this issue from happening going forward.

```css
path {
    fill: var(--vscode-editor-foreground);
}
```

streetcoder123: Looking at it, it seems to to me a good idea. Having said that, I am not a CSS person, so not quite sure. @annielovesjs  @jsondoo Do you have any comments on this?

jcoc611-microsoft: SVG icons that want to have special colors other than the foreground color will still work as they will override the `path` rule. So basically we can remove all the css rules that have `fill: var(--vscode-editor-foreground);` and instead have what I wrote above. For checkmark and stuff like that it will be as it is right now

jcoc611-microsoft: Per discussion we will leave this one as is for now as implementing the project-wide `path` rule breaks a lot of things currently

jcoc611-microsoft (approved):

jsondoo (approved):

jasmingoh (approved):

PR #883, opened 2019-07-11
optimize header more for smaller screens

annielovesjs:
How to test/changes made
- the text in the header will change from Web Template Studio to WTS when the screen gets smaller (< 500px)


src/client/src/containers/Header/index.tsx
```diff
@@ -18,6 +18,7 @@ const Header = () => {
   return (
     <div className={styles.header}>
       <div className={styles.headerTitle}>Web Template Studio</div>
+      <div className={styles.headerTitleSmall}>WTS</div>
```
SahilTara: Would it be possible to call it WebTS instead? Just because there may be confusion with WTS.

jcoc611-microsoft: Also, shouldn't these be internationalized?

annielovesjs: Just changed it to WebTS :)

annielovesjs: @jcoc611-microsoft I don't think it's necessary to have these internationalized just because it should be the same no matter what language it changes to

piotrmark (approved):

jcoc611-microsoft (approved):

jsondoo (approved):

PR #874, opened 2019-07-10
Get Pages through Extension instead of CoreTS directly

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/30277

**Changes**
- Currently, the client directly calls the CoreTS Engine to get pages and put in the wizard Content reducer.
- Now, the client should communicate to the extension that it wants to get pages from CoreTS, extension makes request to the CoreTS and returns the response to the client, which then updates the state.

**How to Check**
- Launch the wizard
- Navigate through it - `Pages` page should be working as usual
- Try to generate a project - Project should be appropriately generated.

PR #891, opened 2019-07-11
Removed client side call to CoreTS and cleaned up code

jasmingoh:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/30717

With #874 and #828, we have removed the client side's `getPages` and `getFramework` calls to CoreTS. The remaining, `getFeature` and `getProjectType`, are not being used at all, so this PR removes all the related code (plus some clean up to get rid of "unused" warnings)

To check, launch the wizard and generate a project. Everything should work just as before.


SahilTara (approved):

piotrmark (approved):
Yay :)

jcoc611-microsoft (approved):

Issue #872, opened 2019-07-10
Create New Project default output path bug


jasmingoh:
Please refer here: https://github.com/microsoft/WebTemplateStudio/pull/861#issuecomment-510138295

Create New Project logics need to take default output path into consideration, and the error box under output path should not be showing by default

![image](https://user-images.githubusercontent.com/20333350/61000421-37a7e680-a312-11e9-9dd2-f85f97169e10.png)

xtinah-w:
Also need to disable input completely and remove user's ability to type their own "Save To" destination

SahilTara:
Think this issue can be closed now

jsondoo:
Closing this - fixed with #893 

> Also need to disable input completely and remove user's ability to type their own "Save To" destination

Fixed with #928

PR #893, opened 2019-07-11
Better reset wizard handling + Output path error message fix

SahilTara:
This PR:
  - Centralizes handling for reset_wizard action. This was done because the code for handling reset_wizard was spread all throughout, and we could possibly miss resetting a member in the state by forgetting do `case WIZARD_INFO_TYPES.RESET_WIZARD: // handle`

  - Fixes odd error message below output path when resetting wizard (#872). This does not handle default output path logic, but it is a step in the right direction!

To test:
- Please build, and then run through the wizard twice (click create new project at end of first run) making sure all your states are properly reset (no weird output path error, and if signed into azure you remain signed in but your selections are wiped)!


src/client/src/reducers/index.ts
```diff
@@ -21,5 +23,29 @@ const rootReducer = combineReducers({
   dependencyInfo
 });
 
-export type AppState = ReturnType<typeof rootReducer>;
+export type AppState = ReturnType<typeof appReducer>;
+
+const rootReducer = (state: AppState | undefined, action: RootAction) => {
+  let passedState: any;
```
SahilTara, line 29: Anyone have any ideas on how to handle the dynamically changing type of the appReducer & passedState here? I don't want to use "any" really.

ngkelly3: I think you'd have to cast the state into its correct type once you figure out what it is, if that's possible.

src/client/src/reducers/index.ts
```diff
@@ -21,5 +23,29 @@ const rootReducer = combineReducers({
...
+
+  if (action.type === WIZARD_INFO_TYPEKEYS.RESET_WIZARD) {
+    const { previewStatus } = state!.wizardContent;
+
+    passedState = {
+      azureProfileData: state!.azureProfileData,
+      dependencyInfo: undefined,
+      generationStatus: undefined,
+      modals: undefined,
```
jsondoo, line 42: Why do we set this to `undefined` rather than using the intial states of each store? I can see that it works but it's unclear to me how it's working

SahilTara: hmm gonna put a comment or something for the future, but essentially think of the store as being undefined right when you launch, Elements that are undefined tell the reducer to replace the item with the initial state. See: https://redux.js.org/recipes/structuring-reducers/initializing-state

image

SahilTara: Notice how we already use the default argument value syntax in every other reducer? This allows us to pass undefined and reset the state.

jsondoo: Okay cool, good to know

src/client/src/reducers/index.ts
SahilTara: Yeah I don't think its worth the effort here really, adds so much overhead to the developer, if they modify the state.

src/client/src/reducers/index.ts
```diff
@@ -21,5 +23,33 @@ const rootReducer = combineReducers({
...
+     */
+    passedState = {
+      azureProfileData: state!.azureProfileData,
+      dependencyInfo: undefined,
+      generationStatus: undefined,
+      modals: undefined,
+      selection: undefined,
+      versions: state!.versions,
+      vscode: state!.vscode,
+      wizardContent: { previewStatus },
```
streetcoder123, line 46: Just for confirmation, correct me if I am saying it wrong. We are assigning `wizardContent` to have `previewStatus` property as defined above. All other attributes of `wizardContent` would be `undefined`, so they will take up the initial state from their respective reducers. Is that right?

src/client/src/reducers/index.ts
```diff
@@ -21,5 +23,33 @@ const rootReducer = combineReducers({
   dependencyInfo
 });
 
-export type AppState = ReturnType<typeof rootReducer>;
+export type AppState = ReturnType<typeof appReducer>;
+
+const rootReducer = (state: AppState | undefined, action: RootAction) => {
+  let passedState: any;
+
+  if (action.type === WIZARD_INFO_TYPEKEYS.RESET_WIZARD) {
```
streetcoder123, line 31: Would it make sense to have the `ResetReducer` as a separate reducer file?

src/client/src/reducers/index.ts
SahilTara: Yup all other attributes will be undefined! Therefore given initial state as you said!

src/client/src/reducers/index.ts
SahilTara: Hmm the problem with that is that it will not apply to the entire root 😦. The point of this is to have the reset handling applied to the entire root state, if that makes sense? I may not be understanding your proposition though, if you think so can you give me an example of how you'd want the code to be?

streetcoder123: I think I get your point. ResetWizard wants to take state from different places and set it to some value but it is not a child of the root reducer. Hence, it needs to be there in parallel with the root reducer. Is that right?

SahilTara: Yup exactly

streetcoder123: Okay, in that case, we can leave it as is for now.

streetcoder123 (approved):

jsondoo (approved):

annielovesjs (approved):

PR #894, opened 2019-07-11
add aria-labels to top nav bar links

annielovesjs:
Changes:
- added aria labels to top nav bar

How to Test with screen reader:
- go to settings -> ease of access -> narrator 
- turn narrator on
- build the extension and run it
- click on the tabs on top nav bar and you should hear "go to ___ page [link]" where [link] is added in by default


SahilTara:
Let me know if this works, I'll approve code looks good. (Can't test this myself at the moment).

piotrmark (approved):

jsondoo (commented):
It seems to work on browser but on extension it always says 'No item in view'

annielovesjs:
> It seems to work on browser but on extension it always says 'No item in view'

I encounter that problem too even with the items that already have aria-label

I find that narrator is pretty inconsistent when you hover over items with aria-labels - sometimes it returns the value of the aria-label and sometimes it saids 'no item in view'

jsondoo (approved):

jcoc611-microsoft:
> > It seems to work on browser but on extension it always says 'No item in view'
> 
> I encounter that problem too even with the items that already have aria-label

After a bit of research, I have found that [Windows Narrator is not good for accessibility.](https://github.com/microsoft/vscode/issues/10729)

Please try one of the following: NVDA, JAWS, WindowEyes, or ZoomText

jcoc611-microsoft (approved):

PR #896, opened 2019-07-11
Removed delete folder function of run-tests

annielovesjs:
Changes:
- removed the portion of the test that deletes the files because it will get automatically deleted in the build pipeline anyways (we'll end up saving about a few minutes)

How to test:
- run yarn template-tests (you can exist as soon as you see yarn install because the test should work since I only deleted the method & created a folder in generate-test)


SahilTara (approved):

SahilTara:
So the build doesn't fail from the test folder?

jcoc611-microsoft (approved):
Works with the caveat that if you want to run it locally you have to manually delete the generated projects

annielovesjs:
> So the build doesn't fail from the test folder?

nope! sibille tested it out

PR #905, opened 2019-07-13
App Services Card in Preview

jsondoo:
**Changes**
- add app services svg
- add an app services card in preview mode (this card should only show up if the wizard is in preview mode)
- remove `useNormalButton` prop and use disabled (disabled buttons should be black and active buttons should always be blue for consistency)

image


annielovesjs (approved):

piotrmark (approved):
Looks good :)

streetcoder123:
Since the `Add Resource` button is not clickable for App Services, should we make it `black` for now so user does not try to click it. Once we have it clickable, we can enable it.

jasmingoh (approved):

jsondoo:
> Since the `Add Resource` button is not clickable for App Services, should we make it `black` for now so user does not try to click it. Once we have it clickable, we can enable it.

Yea, I considered this but I will be making it clickable in my next PR anyways when I have the form ready

PR #908, opened 2019-07-15
Cap Max Number of Pages Generated

### LINKED ISSUE #720: Place limit on how many pages can be added to a product
Currently, users can build what appears to be an infinite amount of pages, we should limit this since the application isn't meant for full-production type applications, but prototyping. Need to find a number to limit it to.

streetcoder123:
Tasks:
- https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/30890
- https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/30985

**Changes**
- Add 20 as the max number of pages allowed to be generated.
- Show a message to the user if he/she tries to add more than 20 pages.
- Abstract UI away from dependency checker component to be reused as a generic component (Notification).

closes #720


jasmingoh:
Not sure if this is intended, but the warning sign looks like it is closer to the top than the bottom border
image

jasmingoh (dismissed):

src/client/src/components/Notification/index.tsx
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  return (
+    <React.Fragment>
+      <div role="img" aria-label={altMessage}>
+        {!showWarning ? (
+          <Checkmark className={styles.iconCheck} />
+        ) : (
+          <Warning className={styles.iconWarning} />
+        )}
```
piotrmark, line 31: NIT: can we revert the condition to skip negation? Would be more readable IMO

src/client/src/containers/SelectOption/index.tsx
```diff
@@ -13,12 +16,36 @@ import { ISelected } from "../../types/selected";
 import { Dispatch } from "redux";
 import RootAction from "../../actions/ActionType";
 
+import { ReactComponent as Warning } from "../../assets/warning.svg";
+import { ReactComponent as Checkmark } from "../../assets/checkgreen.svg";
```
piotrmark: Do we need these imports here?

src/client/src/containers/SelectOption/index.tsx
```diff
@@ -34,28 +61,30 @@ interface ISelectOptionProps {
...
   constructor(props: Props) {
     super(props);
     const { selectedCardIndices } = props;
     this.state = {
-      selectedCardIndices
+      selectedCardIndices,
+      maxPageReached: false,
+      description: props.intl.formatMessage(messages.limitedPages)
     };
+    this.addPage = this.addPage.bind(this);
```
piotrmark: We could avoid this binding by using arrow properties:
https://www.freecodecamp.org/news/react-binding-patterns-5-approaches-for-handling-this-92c651b5af56/

src/client/src/containers/SelectOption/index.tsx
```diff
@@ -263,11 +303,26 @@ class SelectOption extends React.Component<Props, ISelectOptionState> {
...
+      isPagesSelection,
+      intl
     } = this.props;
     return (
       <div>
         <Title>{title}</Title>
+        {isPagesSelection && (
+          <div
+            className={classnames(styles.description, {
+              [styles.borderGreen]: !this.state.maxPageReached,
```
piotrmark: We could also destructure variables from state above

src/client/src/containers/SelectOption/index.tsx
streetcoder123: Nope, left by mistake. :p

src/client/src/containers/SelectOption/index.tsx
streetcoder123: That's good to know!

piotrmark (approved):
That was quick :D Thanks!

jasmingoh (approved):

annielovesjs (approved):

PR #910, opened 2019-07-15
card changes

annielovesjs:
Changes:
- display version number and language type on cards 
- styling changes

How to test:
- should look like this:
pagesDesign2
pagesDesign


piotrmark:
Maybe it would be more readable if we display version first and {xxx} framework below?
Right now it feels a little bit like version applies to JS / Python

src/client/src/translations/en.json
```diff
@@ -140,7 +140,6 @@
   "selectPages.pagesTitleQuestion": "Select pages for your application.",
   "selectPages.webAppTitleQuestion": "Select a project type.",
   "selectableCard.details": "Details",
-  "selectableCard.selected": "Selected",
```
streetcoder123, line 146: Is it not being used now? Did you run `npm run manage:translations`?

annielovesjs: yeah I deleted the text

annielovesjs: since it's not being used anywhere now

annielovesjs:
> Maybe it would be more readable if we display version first and {xxx} framework below?
> Right now it feels a little bit like version applies to JS / Python

Good point! I just swapped the order

src/client/src/components/SelectableCard/index.tsx
```diff
@@ -136,12 +142,6 @@ const SelectableCard = ({
           >
             {(isPagesSelection && <div>{clickCount}</div>) || (
               <div className={styles.selectedText}>
-                <div>
-                  <FormattedMessage
-                    id="selectableCard.selected"
-                    defaultMessage="Selected"
```
jasmingoh, line 144: is removing the "selected" message ok from the accessibility standpoint? will screen reader knows which card is selected without the message?

annielovesjs: The screen reader never picked up the message before I removed it. There are a lot of accessibility problems like this throughout the wizard which I will work on in another PR

jsondoo (approved):
looks good

streetcoder123:
Discussed with @annielovesjs the cards on Azure services page are a little off - slightly bigger for Hosting and smaller for Database. That would be addressed in a separate PR.

streetcoder123 (approved):

src/client/src/components/CardBody/index.tsx
```diff
@@ -5,15 +5,17 @@ import styles from "./styles.module.css";
...
 
 type Props = InjectedIntlProps & IProps;
 
-const CardBody = ({ body, formattedBody, intl }: Props) => {
+const CardBody = ({ body, version, formattedBody, intl }: Props) => {
   return (
     <div className={styles.body}>
-      {body || (formattedBody && intl.formatMessage(formattedBody))}
+      {version && <div>v{version}</div>}
+      {body || (formattedBody && intl.formatMessage(formattedBody))}{" "}
```
piotrmark, line 18: Do we need this space at the end?

src/client/src/components/CardBody/index.tsx
```diff
@@ -5,15 +5,17 @@ import styles from "./styles.module.css";
...
 interface IProps {
   body?: string;
+  version?: string;
   formattedBody?: FormattedMessage.MessageDescriptor;
 }
 
 type Props = InjectedIntlProps & IProps;
 
-const CardBody = ({ body, formattedBody, intl }: Props) => {
+const CardBody = ({ body, version, formattedBody, intl }: Props) => {
```
piotrmark, line 14: I know you didn't change this, but I find this logic with body and formattedBody super confusing. Maybe there is a way to simplify it?

src/client/src/components/SelectableCard/index.tsx
```diff
@@ -102,7 +104,11 @@ const SelectableCard = ({
         )}
         <div className={grid.row}>
           <div className={styles.body}>
-            <CardBody body={body} />
+            {version ? (
```
piotrmark, line 108: Feels like we don't need this condition here as CardBody already checks: {version && ... }

jcoc611-microsoft (approved):

PR #911, opened 2019-07-15
disable top-nav on modal open

### LINKED ISSUE #899: When modal is open, top bar should be disabled
Bug bash revealed users can click and navigate through the top bar when the modal is active. Need to fix.

annielovesjs:
closes #899 
How to test/changes:
- when you open a modal component, you cannot use the top nav bar until you exit

### DIFF
--- a/src/client/src/components/Modal/index.tsx
+++ b/src/client/src/components/Modal/index.tsx
@@ -37,6 +37,7 @@ const getCustomStyles = (MODAL_TYPE: ModalType | undefined) => {
       left: 0,
       right: 0,
       bottom: 0,
+      zIndex: 2000,
       backgroundColor: "rgba(0, 0, 0, 0.6)"
     },
     content: {


src/client/src/components/Modal/index.tsx
```diff
@@ -37,6 +37,7 @@ const getCustomStyles = (MODAL_TYPE: ModalType | undefined) => {
       left: 0,
       right: 0,
       bottom: 0,
+      zIndex: 2000,
```
streetcoder123, line 40: Just out of curiosity, what does this do?

annielovesjs: oh this just gives higher priority to the overlay div so that it overlaps all the other divs (including the header)

streetcoder123: Is there a specific reason for the number `2000` or it could be any number that is greater than other parts of the page?

SahilTara (approved):

src/client/src/components/Modal/index.tsx
piotrmark: It's definitely the later :D Here is a nice strategy example: https://medium.com/science-journal/z-index-organization-in-css-5913fd4c25c9

piotrmark: But the rule of thumb should be: we shouldn't go ridiculously high. 2000 seems to work here ;)

piotrmark (approved):

streetcoder123 (approved):

PR #912, opened 2019-07-15
Make spacebar follow links

ali-kha:
while focused on a link (tabbing through the ui), pressing space will by default act as a "scroll down". This is inconsistent with pressing space on a button which will perform the action. So I changed all the links to also follow on space. If **not** focused on a link, space bar will still act as scroll.


src/client/src/utils/spaceKeyHandler.ts
```diff
@@ -0,0 +1,22 @@
+import * as React from "react";
+import {KEY_EVENTS} from "./constants"
+
+// document.documentElement.addEventListener('keydown', (e) => {
```
SahilTara: Remove commented out code please

src/client/src/utils/spaceKeyHandler.ts
```diff
@@ -0,0 +1,22 @@
...
+//   if (e.key === KEY_EVENTS.SPACE) e.preventDefault();
+// })
+
+const spaceKeyHandler = (e: React.KeyboardEvent<HTMLAnchorElement>) => {
+
+  let anchor = e.target as HTMLAnchorElement;
```
SahilTara: const here please.

src/client/src/utils/spaceKeyHandler.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  //if pressing space on an anchor, prevent page scroll.
+  anchor.onkeydown = (e) => {
```
SahilTara: Curious, does e.stopPropogation() not work?   Wondering why you're attaching an event listener inside the event listener. Feels a bit odd. Can you explain why?

src/client/src/components/Details/index.tsx
```diff
@@ -92,7 +93,7 @@ const Details = ({
             role="button"
             tabIndex={0}
             onClick={handleBackClick}
-            onKeyDown={keyDownHandler}
+            onKeyUp={keyDownHandler}
```
streetcoder123: - Why did you change from `onKeyDown` to `onKeyUp`?
- If we have to change it to `onKeyUp`, we should probably update the function to be called `keyUpHandler` instead.

src/client/src/containers/About/index.tsx
```diff
@@ -38,6 +39,7 @@ const About = ({ versions, intl }: Props) => {
         <a
           className={styles.link}
           href="https://github.com/Microsoft/WebTemplateStudio"
+          onKeyUp={spaceKeyHandler}
```
streetcoder123: Would it make sense to name this function as `keyUpHandler` so that later we can use the same function to handle other key press events as well (if required) using the switch/case statement?

src/client/src/utils/spaceKeyHandler.ts
ali-kha: e.stopPropogation pretty much just "pops" the event bubble, but all event handlers left on the current element (the anchor in this case) will still occur, so that doesn't work, but to be safe I just tested it and it doesn't work. 

The reason I did the inner event listener is because that way the event listener would only be added to elements that we select (whichever elements we add this handler to). The more generic option would be to add an event listener to all a tags with something like `document.getElementsByTageName("a");` I just thought this would be more localized and cause less potential side effects.

src/client/src/components/Details/index.tsx
ali-kha: oh my bad, good catch, this wasn't something I was modifying but I mistook it as my code.

src/client/src/containers/About/index.tsx
ali-kha: yeah I didn't think about that but it would be better, changing that now

annielovesjs (dismissed):
Nice! It works well

src/client/src/containers/Licenses/index.tsx
```diff
@@ -38,7 +40,7 @@ const Licenses = ({
 }: Props) => {
   const LinkRenderer = (props: any) => {
     return (
-      <a href={props.href} className={styles.link}>
+      <a href={props.href} className={styles.link} onKeyUp={keyUpHandler}>
```
streetcoder123: For the Licenses file, we are going to add a `Privacy Modal`. So, these anchor tags would be converted to buttons that would open up the modal on click. Can you revert the changes for this file?

jsondoo (approved):

streetcoder123 (approved):

src/client/src/containers/About/index.tsx
```diff
@@ -38,6 +39,7 @@ const About = ({ versions, intl }: Props) => {
         <a
           className={styles.link}
           href="https://github.com/Microsoft/WebTemplateStudio"
```
markAtMicrosoft, line 41: should this be hard coded into here?  I realize it's not your code but still...  (Consider a new task to fix this, not here)

markAtMicrosoft (approved):

PR #913, opened 2019-07-16
At least 1 page should be generated - Error Message

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/31182

**Changes**
- This PR introduces an error message if the user tries to remove all the pages. This would make more intuitive for the user, why he/she is not able to remove a page.


src/client/src/translations/en.json
```diff
@@ -104,6 +104,7 @@
   "pages.limitedPagesMessage": "You can select up to 20 pages",
   "pages.maxPagesText": "Icon for Max Pages Description",
   "pages.overlimitPagesMessage": "You cannot add more than 20 pages to the project",
+  "pages.noPageGeneration": "You cannot add more than 20 pages to the project",
```
jasmingoh: This doesn't look right.. should the message be the same as https://github.com/microsoft/WebTemplateStudio/pull/913/files#diff-abf0943cba1cc720f993c71a080c2f04R32 ?

streetcoder123: Yeah, I will update it.

jasmingoh (approved):
works!

piotrmark (approved):

jcoc611-microsoft (approved):
:)

Issue #888, opened 2019-07-11
Add App Services


jsondoo:
Add App Services as an Azure service to deploy the generated web app for our users

Part 1 - Creation of App Service instance
- [x] Add App Service Card (#905)
- [x] Add App Service Form (#914)
- [x] Extension validates name and subscription data (#915)
- [x] Send App Service payload from 'Create Project' button to the extension (#959)
- [x] Add App Service to Right Side Nav if created (#968)
- [x] Allow users to only choose one hosting service (Functions or App Service) (#975)
- [x] Tell users they can only choose one hosting service (new string) (#985)
- [x] Warn users about 30 day free trial for Basic Tier ASP in App Service Form (#995)
- [x] Extension creates resource group (#979)
- [x] Extension creates an ASP (#970)
- [x] Extension creates a web app (#1014)
- [x] Make App Service extension a dependency on our extension (#1007)
- [x] Update post gen modal with App Service (#1026)
- [x] Update the Angular package.json to include all required dependencies (#1059)

Part 2 - Deployment of Generated App
- [x] If App Service is selected, generate `settings.json` in the generated project (#1045)
- [x] Update App Service application settings (#1068)
- [x] Update the generated config files to deploy the server folder (#1072)
- [x] Single command for deploying web app (#1077)

Part 3
- [x] Move App Service card out of preview mode (#1078)

xtinah-w:
App Service Form revisions (some changes apply to all Azure forms)
![appServiceForm](https://user-images.githubusercontent.com/24245311/62240599-2dff2500-b38c-11e9-97f4-bc0a5dbe6c70.png)

streetcoder123:
App service is out of preview :)

PR #915, opened 2019-07-16
Add App Service Form

jsondoo:
Issue #888 
Task https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/30983

**Changes**
- App Service Form should open when 'Add Resource` is clicked on the App Service card
- Form data should be able to be saved and retrieved through the `appService` store

**How to Test**
You should be able to test this through the mock vs code API using the **browser** (the extension side commands `EXTENSION_COMMANDS.NAME_APP_SERVICE` and `EXTENSION_COMMANDS.SUBSCRIPTION_DATA_APP_SERVICE` are not in dev yet)
1) Check that you can open App Service Form through the card
2) Check that you can save and retrieve data you put in to the form
3) Check that runtime stack changes based on backend framework chosen

image


src/client/src/containers/AppServiceModal/index.tsx
```diff
@@ -0,0 +1,617 @@
...
+};
+
+interface IStateProps {
+  isModalOpen: boolean;
+  vscode: IVSCodeObject;
+  subscriptionData: ISubscriptionData;
+  subscriptions: [];
+  isValidatingName: boolean;
+  siteNameAvailability: IAvailability;
+  selection: any;
```
jsondoo: Discussed this with @njindal239 and could not figure out a good, simple way to type this (had trouble getting it working with `getAppServiceSelection`). Currently this is also `any` for this for Cosmos and Functions. So it might be better to type this after doing some refactoring for all the forms. this If you guys have suggestions let me know.

jsondoo:
There's also a lot of inconsistencies in the code for Azure Services so here are some things we can clean up in separate PRs
- Rename `cosmosDbModalData` -> `azureServicesInitialStateData`
- Move all azure services messages to `azureServiceOptions` (and rename `mockData` to `data`)
- clean up `selections` array to be just `selections` (single object) for Cosmos/Functions
- clean up `classname` vs `className` (we use a mix of both)
- consistent command names

src/client/src/reducers/wizardSelectionReducers/services/appServiceReducer.ts
```diff
@@ -0,0 +1,95 @@
...
+  siteNameAvailability: IAvailability;
+  selection: ISelectedAppService | null;
+  wizardContent: IServiceContent;
+  chooseExistingRadioButtonSelected: boolean;
+}
+
+const initialState: IAppServiceSelection = {
+  siteNameAvailability: {
+    isSiteNameAvailable: false,
+    message: "App name unavailable"
```
markAtMicrosoft, line 50: resource string?

markAtMicrosoft (changes requested):

src/client/src/containers/AppServiceModal/index.tsx
Tanya0609: what is the variable "selection" for?

src/client/src/reducers/wizardSelectionReducers/services/appServiceReducer.ts
jsondoo: Okay I looked into this, this should have been a simple fix but changing this from a `string` to a `FormattedMessage.MessageDescriptor` requires lot more changes then I expected.

The action type for setting availability is coupled in `setAccountAvailability.ts` so we need to refactor all three forms:
1) changing to use intl.formatMessage, rather than this message
2) changing how we update this state
3) changing button validation 

After discussing with Tanya and Nipun, the best solution would be to just remove this message field from the reducer `siteNameAvailability`, and just keep the boolean, and then in the component render the resource string. 

Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/31327

src/client/src/selectors/appServiceSelector.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @param services An object of all the services available in Project Acorn
+ */
+const getAppServiceSelectionInDropdownForm = (appState: any): any => {
+  const selection = appState.selection.services.appService.selection;
+  if (!_.isEmpty(selection)) {
+    const selectionInformation: ISelectionInformation = {
+      dropdownSelection: {},
+      previousFormData: selection
+    };
+    for (const selectionKey in selection) {
```
Tanya0609: NIT: tslint is complaining, it should use Object.keys to access selection rather than a for loop.

so, maybe something like this will work? Then the types can be set too.
      selectionInformation.dropdownSelection.internalName = {
        label: selection.internalName,
        value: selection.internalName
      };

Tanya0609 (changes requested):

markAtMicrosoft (dismissed):

src/client/src/containers/AppServiceModal/index.tsx
jsondoo: This is typed now.

`selection` stores the form data in the appService store

Tanya0609 (approved):

jasmingoh (approved):

markAtMicrosoft (approved):

Issue #767, opened 2019-06-20
can't launch


hudyyy:
Repro steps:

1.install 
2.command+shift+p , Web Template Studio: Launch

Action: Sync-Engine
Error type: Error
Error Message: Could not sync to template repository

Version: 0.0.1915901
OS: darwin
VS Code Version: 1.35.1

<details>
<summary>Call Stack</summary>

```
LaunchExperience.<anonymous> (/Users/<user>/.vscode/extensions/wasteamaccount.webtemplatestudio-dev-nightly-0.0.1915901/out/launchExperience.js:40:23)
Generator.next (<anonymous>)
fulfilled (/Users/huangqingyu/.vscode/extensions/wasteamaccount.webtemplatestudio-dev-nightly-0.0.1915901/out/launchExperience.js:4:58)
```

</details>

Tanya0609:
Hi, @hudyyy, thank you for reporting this issue! We will take a look at it and get back to you soon.

jasmingoh:
Hi @hudyyy, I'm one of the developer working on the extension. We are aware of this issue, but unfortunately, we are not able to reproduce it. It would be awesome if you can provide us with more details about what you did to run into that error.

We have a new version of the extension up on VS Code market place now, 0.0.1918291. Please re-launch the extension and let us know if you're still encountering the problem.

SahilTara:
@jasmingoh Have you tried provisioning a vm with low resources (4gb ram limit). Have you tried trying it on a computer with lower specs (since the garage laptops are very nice).

jasmingoh:
@SahilTara I'm able to reproduce it now using one of our mac

jasmingoh:
@hudyyy We have a solution in mind now, and we are working on it. Will let you know when there's update.
ADO task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_sprints/taskboard/WTS%20Team/Intern%20GitHub/Y19-VS/08

SahilTara:
@jasmingoh only people with access to your sprint board can see that link.

vvatkov:
@jasmingoh I receive that error message when I launch the extension on Windows. The extension, however, is built on my Windows development environment using the provided **createVsix** installation script.

Tanya0609:
Hi @vvatkov, @hudyyy, the solution to this issue has been merged into our code base, and will be released next Monday. Please download the new release extension from marketplace or update your current extension, and the issue should be fixed for you. 

[...]

PR #916, opened 2019-07-16
re-architecture: extension call corets-cli

### LINKED ISSUE #955: cannot launch, "Cannot read property 'value' of undefined
when i run the `Web Template Studio: Launch...` command I get the following output:

image

to repro:

browse to extension in the marketplace
click on "install"
VS code (stable) opens, install extension
press `F1` to open the command palette
choose `Web Template Studio: Launch" command

if I reload VS Code and try again, I get the same set of errors in the Output but i do see the editor open inside VS code.

Tanya0609:
closes #955

**Summary to this change:**

1. put corets cli into build file, so run ./build, it builds the cli executable 
2. use a child process to start the corets cli executable 
3. send commands to cli and receive responses
4. interpret responses and return to UI
5. clean up/remove the code related to corets-api, including all the dlls

There are 6 commands in total:
1. sync command
2. get frameworks
3. get pages
4. generate

5. get features,it returns available Azure services based on the front-end and back-end frameworks ( not used in WebTS atm)
6. get project types ( not used in WebTS atm )

**How to test:**
1. switch to my branch and do git submodule update
2. do ./build
3. start the extension, and generate a project (with and without Azure services) you should be able to generate a project 
4. in post generation modal, click "create a new project", you should be able to generate another project
5. in post generation modal, click "close wizard" and do "ctrl+shift+p" to start a new wizard, you should be able to generate another project.

**Make sure to run:**
git submodule init
git submodule update
___

**Test Cases:**
- [ ] you are able to generate a project with Azure Services 
- [ ] you are able to generate a project without Azure Services 
- [ ] in post generation modal, click "close wizard" and do "ctrl+shift+p" to start a new wizard, you should be able to generate another project.
- [ ] in post generation modal, click "create a new project", you should be able to generate another project


jsondoo:
@Tanya0609 so will we be merging this after redmond?

Tanya0609:
It depends on how stable this change is. If not stable, then we should not merge it before Redmond, 

The thing is coreTS (this PR) always works and returns data, but sometimes the reactPanel will just be gone and all the views will be vanished. There will be more investigation on this issue.

Edit: This issue has been fixed :) It is caused by dependency checker loads before get frameworks load. It wasn't happen before because get frameworks will take a while to load. But since we are calling coreTS cli, the performance has been improved and get frameworks takes less time to load, this turns to an issue.

I updated my PR with fix to this issue. Now it is stable and are safe to merge anytime when it gets approved 👍

src/extension/src/coreTemplateStudio.ts
```diff
@@ -49,40 +47,32 @@ export class CoreTemplateStudio {
...
 
-    const port = await portfinder.getPortPromise({
-      port: CONSTANTS.START_PORT
+    const spawnedProcessCli = spawn(cliPath, [], {
+      cwd: cliWorkingDirectory
     });
 
-    let spawnedProcess = execFile(
-      `${apiPath}`,
-      [`--urls=http://localhost:${port}`],
```
jcoc611-microsoft: Should be `[]`

jcoc611-microsoft (changes requested):
Waiting for CoreTS update fixing bug with generated project output directory

src/extension/src/coreTemplateStudio.ts
```diff
@@ -49,40 +47,32 @@ export class CoreTemplateStudio {
...
+    if (os.platform() !== CONSTANTS.CLI.WINDOWS_PLATFORM_VERSION) {
       // Not unsafe as the parameter comes from trusted source
       // tslint:disable-next-line:non-literal-fs-path
-      fs.chmodSync(apiPath, 0o755);
+      fs.chmodSync(cliPath, 0o755);
     }
 
-    const port = await portfinder.getPortPromise({
-      port: CONSTANTS.START_PORT
+    const spawnedProcessCli = spawn(cliPath, [], {
```
SahilTara, line 70: Curious does this work on mac, with the cwd? I remember I had problems with getting spawn to keep the cwd on mac. Which caused the relative paths passed in to not work. Can you just make sure of this (Try both local dev host and vsix). Might have just been an API thing though, or node fixed the problem.

jcoc611-microsoft: I really hope it works because we need the spawn shell to redirect stdio

src/extension/src/coreTemplateStudio.ts
```diff
@@ -1,16 +1,16 @@
...
 import { CONSTANTS } from "./constants";
-import { GenerateCommand } from "./signalr-api-module/generateCommand";
-import { SyncCommand } from "./signalr-api-module/syncCommand";
-import { ICommandPayload } from "./signalr-api-module/commandPayload";
+import { ICommandPayload } from "./types/commandPayload";
+import { IGenerationPayloadType } from "./types/generationPayloadType";
+import { EventEmitter } from "events";
+import { IEngineGenerationPayloadType } from "./types/engineGenerationPayloadType";
+
+class CliEventEmitter extends EventEmitter { }
```
DIGarnier, line 13: Considering that `CliEventEmitter `is in no way different from `EventEmitter `, what was the reason to create a new type here?

src/extension/src/coreTemplateStudio.ts
```diff
@@ -94,144 +84,143 @@ export class CoreTemplateStudio {
...
   }
 
-  public async getFrameworks(projectType: string): Promise<any> {
-    const url = new URL(CONSTANTS.API.ENDPOINTS.FRAMEWORK, this._url);
-    url.searchParams.append(
-      CONSTANTS.API.QUERY_PARAMS.PROJECT_TYPE,
-      projectType
-    );
+  public async sync(payload: ICommandPayload): Promise<any> {
+    let syncCommand = `${CONSTANTS.CLI.SYNC_COMMAND_PREFIX} -p ${payload.payload.path}\n`;
```
DIGarnier: Given that syncCommand is not modified after initialization this could be `const`

src/extension/src/coreTemplateStudio.ts
```diff
@@ -94,144 +84,143 @@ export class CoreTemplateStudio {
...
+    this._processCli.stdin.write(syncCommand);
+    this.cliEvents.on(CONSTANTS.CLI.SYNC_PROGRESS_STATE, (data) => {
+      payload.liveMessageHandler(data['status'], data['progress']);
+    });
+    return await this.awaitCliEvent(CONSTANTS.CLI.SYNC_COMPLETE_STATE);
```
DIGarnier: Because you are returning a promise `await` is not needed here

src/extension/src/coreTemplateStudio.ts
```diff
@@ -94,144 +84,143 @@ export class CoreTemplateStudio {
...
-      .then((response: Response) => {
-        return response.json();
-      })
-      .catch((error: Error) => {
-        throw Error(error.toString());
-      });
+  public async getFrameworks(projectType: string): Promise<any> {
+    const getFrameworksCommand = `${CONSTANTS.CLI.GET_FRAMEWORKS_COMMAND_PREFIX} -p ${projectType}\n`;
+    this._processCli.stdin.write(getFrameworksCommand);
+    return await this.awaitCliEvent(CONSTANTS.CLI.GET_FRAMEWORKS_COMPLETE_STATE);
```
DIGarnier: Same as previous comment

src/extension/src/coreTemplateStudio.ts
```diff
@@ -94,144 +84,143 @@ export class CoreTemplateStudio {
...
-    return await fetch(url.href, { method: CONSTANTS.API.METHODS.GET })
-      .then((response: Response) => {
-        return response.json();
-      })
-      .catch((error: Error) => {
-        throw Error(error.toString());
-      });
+    const getPagesCommand = `${CONSTANTS.CLI.GET_PAGES_COMMAND_PREFIX} -p ${projectType} -f ${frontendFramework} -b ${backendFramework}\n`;
+    this._processCli.stdin.write(getPagesCommand);
+    return await this.awaitCliEvent(CONSTANTS.CLI.GET_PAGES_COMPLETE_STATE);
```
DIGarnier: Same as previous comment concerning `await`

src/extension/src/coreTemplateStudio.ts
```diff
@@ -94,144 +84,143 @@ export class CoreTemplateStudio {
...
 
-  public async sync(payload: ICommandPayload): Promise<any> {
-    let connection = await this.connectToCoreApiHub();
-    let command = new SyncCommand(payload);
-    return await command.execute(connection);
+  public async getProjectTypes(): Promise<any> {
+    // to use this in client
+    const getProjectTypesCommand = `${CONSTANTS.CLI.GET_PROJECT_TYPES_COMMAND_PREFIX}\n`;
+    this._processCli.stdin.write(getProjectTypesCommand);
+    return await this.awaitCliEvent(CONSTANTS.CLI.GET_PROJECT_TYPES_COMPLETE_STATE);
```
DIGarnier: Same as previous comment concerning `await`

src/extension/src/coreTemplateStudio.ts
```diff
@@ -94,144 +84,143 @@ export class CoreTemplateStudio {
...
   }
 
   public async generate(payload: ICommandPayload): Promise<any> {
-    let connection = await this.connectToCoreApiHub();
-    let command = new GenerateCommand(payload);
-    return await command.execute(connection);
+    var generatePayload = JSON.stringify(this.makeEngineGenerationPayload(<IGenerationPayloadType>(payload.payload)));
```
DIGarnier: Should be const as we don't modify the variable

src/extension/src/coreTemplateStudio.ts
```diff
@@ -94,144 +84,143 @@ export class CoreTemplateStudio {
...
+    const generateCommand = `${CONSTANTS.CLI.GENERATE_COMMAND_PREFIX} -d ${generatePayload}\n`;
+    this._processCli.stdin.write(generateCommand);
+    this.cliEvents.on(CONSTANTS.CLI.GENERATE_PROGRESS_STATE, (data) => {
+      payload.liveMessageHandler(data);
+    });
+    return await this.awaitCliEvent(CONSTANTS.CLI.GENERATE_COMPLETE_STATE);
```
DIGarnier, line 183: Same as previous comment concerning `await`

src/extension/src/coreTemplateStudio.ts
```diff
@@ -94,144 +84,143 @@ export class CoreTemplateStudio {
...
+  }
+
+  private makeEngineGenerationPayload(
+    payload: IGenerationPayloadType
+  ): IEngineGenerationPayloadType {
+    let {
```
DIGarnier: Should be const as we don't modify the variable

DIGarnier (commented):
Small nits!

src/extension/src/coreTemplateStudio.ts
jcoc611-microsoft: Didn't work using `EventEmitter` directly for some reason

src/extension/src/coreTemplateStudio.ts
Tanya0609: good catch! fixed in my PR :)

src/extension/src/coreTemplateStudio.ts
Tanya0609: The reason is I want promise get resolved before it gets returned. Thus I use await here. Await puts an pause in the code.

src/extension/src/coreTemplateStudio.ts
Tanya0609: fixed in my PR :)

src/extension/src/coreTemplateStudio.ts
Tanya0609: good catch! updated in my PR :)

src/extension/src/coreTemplateStudio.ts
Tanya0609: "await suspends the execution until an asynchronous function return promise is fulfilled and unwraps the value from the Promise returned. " from Typescipt docs

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-1-7.html

src/extension/src/coreTemplateStudio.ts
SahilTara: If no one has time to verify this right now, I can check this in ~48 hours (exams almost over 😄)

src/extension/src/coreTemplateStudio.ts
```diff
@@ -40,7 +38,7 @@ export class CoreTemplateStudio {
     }
 
     let platform = process.platform;
```
jasmingoh: `const` here since platform is not being modified after declaration

src/extension/src/coreTemplateStudio.ts
```diff
@@ -94,144 +84,151 @@ export class CoreTemplateStudio {
...
 
-  public getPort(): number {
-    return this._port;
+  // This function is a listener, in the constructor, it gets attached
+  // then it will always get triggered when a command is write to the cli and there is responses from cli, until the process gets killed
+  public async readStream(process: ChildProcess) {
+    let data = "";
+    process.stdout.on("data", (chunk) => {
+      data += chunk;
+      var responses = data.toString().split("\n");
```
jasmingoh: use const instead of var

src/extension/src/coreTemplateStudio.ts
```diff
@@ -94,144 +84,151 @@ export class CoreTemplateStudio {
...
+      for (let i = 0; i < responses.length - 1; i++) {
+        var result = JSON.parse(responses[i]);
```
jasmingoh: use const instead of var

jsondoo (commented):
Hey Tanya, after installing .NET, we are getting this error when we run `./build`
```
warning __EXTENSIONNAME__@0.0.0-UNTRACKEDVERSION: The engine "vscode" appears to be invalid.
$ gulp build
[16:03:49] Using gulpfile ~\Desktop\WebTemplateStudio\src\extension\gulpfile.js
[16:03:49] Starting 'build'...
[16:03:49] Starting 'clean'...
[16:03:49] Finished 'clean' after 40 ms
[16:03:49] Starting 'internal-nls-compile'...
[16:03:54] [i18n] Generating localized messages for 'en' resulted in the following problems:
[16:03:54] [i18n]
[16:03:54] [i18n] No localized message found for key info.startingServerMessage
[16:03:54] [i18n] No localized message found for key dialog.logoutPrompt
[16:03:54] [i18n]
[16:03:54] [i18n]
[16:03:55] Finished 'internal-nls-compile' after 5.9 s
[16:03:55] Starting 'add-locales'...
[16:03:55] Finished 'add-locales' after 20 ms
[16:03:55] Starting '<anonymous>'...
[16:03:55] Finished '<anonymous>' after 1.37 ms
[16:03:55] Finished 'build' after 5.97 s
Done in 8.69s.
Microsoft (R) Build Engine version 15.9.20+g88f5fadfbe for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

MSBUILD : error MSB1009: Project file does not exist.
Switch: src/CoreTemplateStudio/code/src/CoreTemplateStudio/CoreTemplateStudio.Cli/CoreTemplateStudio.Cli.csproj
Microsoft (R) Build Engine version 15.9.20+g88f5fadfbe for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

MSBUILD : error MSB1009: Project file does not exist.
Switch: src/CoreTemplateStudio/code/src/CoreTemplateStudio/CoreTemplateStudio.Cli/CoreTemplateStudio.Cli.csproj
Microsoft (R) Build Engine version 15.9.20+g88f5fadfbe for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

MSBUILD : error MSB1009: Project file does not exist.
Switch: src/CoreTemplateStudio/code/src/CoreTemplateStudio/CoreTemplateStudio.Cli/CoreTemplateStudio.Cli.csproj
```

src/extension/src/coreTemplateStudio.ts
Tanya0609: In the original API call, they also use cwd for all platforms. 

I tested my code on Mac, it also works!

src/extension/src/coreTemplateStudio.ts
DIGarnier: You're still returning a promise, it has the same exact behavior if you remove the await, I would incite you to test it!

src/extension/src/coreTemplateStudio.ts
```diff
@@ -94,144 +84,151 @@ export class CoreTemplateStudio {
...
-      CONSTANTS.API.QUERY_PARAMS.PROJECT_TYPE,
-      projectType
-    );
+  public async sync(payload: ICommandPayload): Promise<any> {
+    const syncCommand = `${CONSTANTS.CLI.SYNC_COMMAND_PREFIX} -p ${payload.payload.path}\n`;
+    this._processCli.stdin.write(syncCommand);
+    this.cliEvents.on(CONSTANTS.CLI.SYNC_PROGRESS_STATE, (data) => {
+      payload.liveMessageHandler(data['status'], data['progress']);
+    });
+    return await this.awaitCliEvent(CONSTANTS.CLI.SYNC_COMPLETE_STATE);
```
xWiiLLz: A note on what your answer was to David:
> The reason is I want promise get resolved before it gets returned. Thus I use await here. Await puts an pause in the code.

The only _pause_ that would happen here would be before resolving the promise to the caller of the method. In the case where you're returning a promise, if you `await` for it to resolve before returning it (and thus resolving it), it adds one layer more than needed to do **exactly the same thing**.

Now, this is a common mistake when writing asynchronous functions in TS/JS, which is why there is an [ESLint rule for it that is well documented](https://eslint.org/docs/rules/no-return-await). If you are still unsure about the behavior, I would recommend [reading this short article](https://jakearchibald.com/2017/await-vs-return-vs-return-await/) that explains it more thoroughly. 😊

xWiiLLz (changes requested):

Tanya0609:
@jsondoo This is caused by Git Submodule files aren't inside this repo. Thus, "Project file does not exist."

Try: 
**git submodule init
git submodule update**

This should solve this error.

src/extension/src/coreTemplateStudio.ts
abelfodil: If you look at `src/extension/node_modules/@types/node/index.d.ts` line 938, you can see `EventEmitter` defined and exported as a class.

Have you tried defining `private cliEvents: EventEmitter;` and using `this.cliEvents = new EventEmitter();` in your constructor?

If so, could you please post your error?

src/extension/src/coreTemplateStudio.ts
```diff
@@ -94,144 +84,151 @@ export class CoreTemplateStudio {
...
-      })
-      .catch((error: Error) => {
-        throw Error(error.toString());
+  private async awaitCliEvent(eventName: string): Promise<any> {
+    return new Promise((resolve, reject) => {
+      this.cliEvents.once(eventName, (data) => {
+        this.cliEvents.removeAllListeners();
+        resolve(data);
+      });
+      this.cliEvents.once("error", (data) => {
```
abelfodil: Suppose your have two events `X` and `Y` and you subscribe to both of them. If any operation fails and sends the `error` event, all listeners to both events will be removed even though it could be that none of them failed.

Furthermore, if either event succeeds, it will stop listening to the other event.

Is this the intended behavior? From what I read from the older code, for each operation, there is an associated failure (instead of a global one).

abelfodil: As an additional note (this time on style), `EventEmitter.once` return `this`, so you can chain calls to once like this: `this.cliEvents.once(...).once(...)`

src/extension/src/coreTemplateStudio.ts
```diff
@@ -94,144 +84,151 @@ export class CoreTemplateStudio {
...
+      data += chunk;
+      var responses = data.toString().split("\n");
+      for (let i = 0; i < responses.length - 1; i++) {
+        var result = JSON.parse(responses[i]);
+        this.cliEvents.emit(result['messageType'], result['content']);
+      }
+      data = responses[responses.length - 1];
+    });
+
+    process.stderr.on("data", (data) => {
```
abelfodil, line 107: According to your comment 

> it will always get triggered when a command is write to the cli and there is responses from cli, until the process gets killed

It seems like this function relies on the fact that the child process will *always* write on `stderr` before exiting.

I would recommend using `process.on('exit', callback)` function instead.

src/extension/src/coreTemplateStudio.ts
Tanya0609: This is what I read from [JavaScript documents ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await)about **await**: 

> The await expression causes async function execution to pause until a Promise is settled, that is fulfilled or rejected, and to resume execution of the async function after fulfillment. When resumed, the value of the await expression is that of the fulfilled Promise.

In the article you linked, it gives better insights about promise/await. I guess because I am using **return** a promise here rather than a promise, **return** has the logic to wait and resolve a promise. Thus, I could emit the **await** in my code. 

It also points out **return** a promise will resolve the defer the result, while **return await** a promise will turn its rejection into a throw statement, which will be useful if there is a try catch.

Thank you so much for sharing this resource and will update my PR accordingly :)

src/extension/src/coreTemplateStudio.ts
SahilTara: Cool! just wanted to make sure since I remember having to use exec, guess its been fixed! (Or that it was an older mac version problem).

jsondoo:
@Tanya0609 ok great, running `git submodule init` gets rid of the error, thanks

src/extension/src/coreTemplateStudio.ts
jcoc611-microsoft: > Is this the intended behavior? From what I read from the older code, for each operation, there is an associated failure (instead of a global one).


Since the CLI works synchronously, there is only one command running at a time (so only one event other than error being subscribed). I think because of this, it should work.

src/extension/src/coreTemplateStudio.ts
jcoc611-microsoft: Well technically, the CLI should *never* exit (unless killed on `.stop()`). Of course, things can go wrong and cause the CLI to crash unexpectedly. Thus, maybe it's better if we handle both cases.

jasmingoh:
Great work @Tanya0609 - this PR fixed the launching issue on MacOS (#767, #955)

src/extension/src/coreTemplateStudio.ts
Tanya0609: updated in my PR :)

src/extension/src/coreTemplateStudio.ts
Tanya0609: updated in my PR

src/extension/src/coreTemplateStudio.ts
Tanya0609: Great Catch! 

Will something like this handle both cases?
`    try {
      process.stderr.on("data", (data) => {
        this.cliEvents.emit("error", data.toString());
      });
    } catch (error) {
      process.on("exit", (error) => {
        this.cliEvents.emit("error", error.toString());
      });
    }`

abelfodil: According to Node.js documentation https://nodejs.org/api/process.html#process_event_exit, I would rather use something like this:

``` typescript
process.stderr.on("data", (data) => this.cliEvents.emit("error", data.toString()));

process.on("exit", (code) => this.cliEvents.removeAllListeners());
// or 
process.on("exit", (code) => this.cliEvents.emit("error", `process exited with code ${code}`));
```

Tanya0609: Will update PR accordingly :)

src/extension/src/coreTemplateStudio.ts
Tanya0609: Thank you for pointing out the potential concerns and style suggestion. As Juan explained, because the way the Cli setup, we are good on the case you pointed. I updated my PR with your style suggestion 👍

jcoc611-microsoft (approved):
💯

jsondoo (approved):
Tested out azure services and various wizard things on Windows - looks good :)

jasmingoh (approved):

PR #923, opened 2019-07-17
Warn the User on Azure Logout

### LINKED ISSUE #901: Users should receive a warning before they sign out of Azure
There is currently no user feedback if a user accidentally clicks sign out in the Azure page.

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/31315

**Changes**
- This PR introduces a warning from the extension (lower right corner) when the user clicks on 'Sign Out' of Azure. This is to confirm whether he/she wants to sign out so that he/she does not accidentally get logged out of Azure.

closes #901


src/client/src/translations/whitelist_en.json
```diff
@@ -93,9 +92,9 @@
   "selectFrontendFramework.selectFrontendFramework",
   "pages.limitedPagesMessage",
   "pages.overlimitPagesMessage",
+  "pages.noPageGeneration",
   "pages.maxPagesText",
   "selectPages.pagesTitleQuestion",
-  "selectPages.webAppTitleQuestion",
```
annielovesjs, line 99: why did you remove this?

src/extension/src/constants.ts
```diff
@@ -327,6 +327,10 @@ export namespace DialogMessages {
     "dialog.resetPagesPrompt",
     "Switching Frameworks will reset pages in your queue. Are you sure you want to proceed?"
   );
+  export const logoutPrompt: string = localize(
+    "dialog.logoutPrompt",
+    "Are you sure you want to Sign Out of your Azure Account?"
```
annielovesjs: make sign out and account lower case

src/client/src/translations/whitelist_en.json
streetcoder123: I did not. I ran `npm run manage:translations` as well but I don't know why it still shows up. I don't know if it is being used anywhere.

jasmingoh (dismissed):

src/extension/src/azure/azure-auth/azureAuth.ts
```diff
@@ -46,16 +46,35 @@ export abstract class AzureAuth {
...
         "ms-vscode.azure-account"
       )!.exports;
     }
   }
 
+  public static async promptUsersToLogout() {
+    return await vscode.window
+      .showInformationMessage(
+        DialogMessages.logoutPrompt,
+        ...[DialogResponses.yes, DialogResponses.no]
```
piotrmark, line 59: We spread array here? Is it different than just passing these 2 arguments without array? :)

src/extension/src/azure/azure-auth/azureAuth.ts
```diff
@@ -66,15 +85,20 @@ export abstract class AzureAuth {
...
-        throw new AuthorizationError(CONSTANTS.ERRORS.LOGOUT_FAILED);
+        await vscode.commands.executeCommand("azure-account.logout");
+        // Make sure it did not return from timeout
+        if (this.api.status === CONSTANTS.AZURE_LOGIN_STATUS.LOGGED_IN) {
+          throw new AuthorizationError(CONSTANTS.ERRORS.LOGOUT_FAILED);
+        }
       }
+      return true;
+    } else {
+      return false;
```
piotrmark: Can we reduce nesting by starting with `if (!userConfirmation.payload.signOut) {return false}`?

src/extension/src/azure/azure-auth/azureAuth.ts
streetcoder123: I am not entirely sure. This is the definition of the `showInformationMessage` function:
`export function showInformationMessage<T extends MessageItem>(message: string, ...items: T[]): Thenable<T | undefined>;`

I did it the same way it is done for reset pages.

src/extension/src/azure/azure-auth/azureAuth.ts
streetcoder123: Yup, makes sense :)

src/extension/src/azure/azure-auth/azureAuth.ts
jcoc611-microsoft: The definition is fine, but I think you can still pass arguments directly as Piotr suggested. In the function, the arguments after `message` will be pushed into an array `items`

piotrmark: Yes, so it's the 'rest parameters', it basically says 'gimme as many parameters as you want and I'll make an array from them' (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters). In the function call we declare the array and spread it, so we're just passing 2 arguments (it's equivalent to `showInformationMessage(DialogMessages.logoutPrompt, DialogResponses.yes, DialogResponses.no)`).
But, if you say it's consistent with the rest of the pages, let it be ;)

piotrmark (dismissed):

src/extension/src/azure/azure-auth/azureAuth.ts
```diff
@@ -46,16 +46,35 @@ export abstract class AzureAuth {
...
+      .showInformationMessage(
+        DialogMessages.logoutPrompt,
+        ...[DialogResponses.yes, DialogResponses.no]
+      )
+      .then((selection: vscode.MessageItem | undefined) => {
+        let userConfirmation = {
+          signOut: false
+        };
+        if (selection === DialogResponses.yes) {
+          userConfirmation = {
```
jcoc611-microsoft: Nit `userConfirmation.signOut = true`

src/extension/src/azure/azure-auth/azureAuth.ts
streetcoder123: I see. It makes sense to me now. I completely agree we can do it that way and in fact I can change it at both places. But I feel, that the way it is being processed in the function, this is more intuitive as the options (`yes` and `no`) are passed into the function.
It is kind of weird in general practice to make an array and spread it right away (lol) but I think here it makes sense to show the purpose of those arguments to the reader.

jasmingoh (approved):

piotrmark (approved):

annielovesjs (approved):

PR #926, opened 2019-07-17
Fixed issue around project name limit

ali-kha:
Project name limit used to give an error when the 50th char was entered, and would block from continuing. This PR fixes that so that when the 50th key is entered, nothing happens, but when the user attempts to enter a 51st key, an error will pop up and the name will not change (it will block the new char). Also this error will no longer block the next button. Not blocking the next button is the reason why the existing redux store could not be used as that state would also block the next button.


src/client/src/containers/ProjectNameAndOutput/index.tsx
```diff
@@ -60,13 +60,22 @@ const messages = defineMessages({
     id: "projectName.ariaProjectName",
     defaultMessage: "Project Name Input"
   },
+  nameTooLong: {
+    id: "projectNameError.nameTooLong",
+    defaultMessage:
+      "Project name can only be " +
+      PROJECT_NAME_CHARACTER_LIMIT +
+      " characters long"
```
jsondoo: I know you didn't write this but we should change this to "Project name can only be {characterLimit} characters long" and pass ` PROJECT_NAME_CHARACTER_LIMIT` when you format the message (in this PR or another one)

ali-kha: fixed it

src/client/src/components/Input/index.tsx
```diff
@@ -5,6 +5,7 @@ import styles from "./styles.module.css";
 
 interface IProps {
   handleChange: (e: React.SyntheticEvent<HTMLInputElement>) => any;
+  handleKeyDown?: (e: React.KeyboardEvent<HTMLInputElement>) => any;
```
annielovesjs: might want to set to void since you're not returning anything from the function

src/client/src/containers/ProjectNameAndOutput/index.tsx
```diff
@@ -98,6 +106,18 @@ const ProjectNameAndOutput = (props: Props) => {
...
     updateProjectName(element.value);
   };
+  const validateKey = (e: React.KeyboardEvent<HTMLInputElement>) => {
+      const element = e.target as HTMLInputElement;
+      const inputKeyCheck = /^[A-Za-z0-9_\- ]$/
+
+      if(element.value.length === 50 && inputKeyCheck.test(e.key)){
+        setprojectNameMaxLength(true);
+        e.stopPropagation();
+      }else{
```
jsondoo: We should use Prettier to format these files

src/client/src/components/Input/index.tsx
ali-kha: done

src/client/src/containers/ProjectNameAndOutput/index.tsx
ali-kha: woops I had it turned off

src/client/src/containers/ProjectNameAndOutput/index.tsx
```diff
@@ -60,13 +60,20 @@ const messages = defineMessages({
...
+      `Project name can only be {maxLength} characters long`
+  },
   outputPathTitle: {
     id: "projectName.outputPathTitle",
     defaultMessage: "Save To"
   }
 });
 
 const ProjectNameAndOutput = (props: Props) => {
+  const [projectNameMaxLength, setprojectNameMaxLength] = React.useState(false);
```
jsondoo: NIT: `setProjectNameMaxLength`

jsondoo (dismissed):
looks good

streetcoder123 (dismissed):

piotrmark (dismissed):

piotrmark (approved):

jsondoo (approved):

streetcoder123 (approved):

PR #927, opened 2019-07-17
Change azure page design + copy 

annielovesjs:
Changes:
- styling changes 
- made it more responsive
- changed text

How to test:
- Should look like this on normal screen (on extension): 
azurechanges

- should look like this on large screen (this is on localhost)
largesscreenazure


streetcoder123 (commented):
The width of the right sidebar seems to increase a little bit as we go from pages page to the azure services page. Maybe that has something to do with the added scrollbar in the azure services page. Can we do something about it?

markAtMicrosoft:
I thnk that subtle change in the width of the sidebar really needs to get resolved.  Either that or it's a bug and we need to see a link to the bug and commitment to resolve.

markAtMicrosoft (commented):
I thnk that subtle change in the width of the sidebar really needs to get resolved. Either that or it's a bug and we need to see a link to the bug and commitment to resolve.  Unfortunately the screen shots make it look the same but it's not.
image
image

annielovesjs:
@markAtMicrosoft @njindal239 it's actually a bug that's on the dev branch as well. I will log this issue and address it in a separate pr shortly

jasmingoh (approved):

streetcoder123 (approved):

xtinah-w (approved):

PR #936, opened 2019-07-18
Reset all pages feature

annielovesjs:
Changes:
- on the side bar, you can reset all your pages instead of manually delete each one of them 

How to test:
- select some pages and then hit the reset button/icon and it should erase all your page selections (except for one blank page)
- looks like this:
reset


piotrmark (dismissed):

streetcoder123:
The reset pages icon is not visible on VSCode light mode. I think you need to add VSCode variable css property for its color :)

image

src/client/src/assets/i-close.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="12" height="12" viewBox="0 0 12 12" fill="none" xmlns="http://www.w3.org/2000/svg">
```
streetcoder123: Should we give an alt tag to this svg?

src/client/src/assets/i-hamburger.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="20" height="16" viewBox="0 0 20 16" fill="none" xmlns="http://www.w3.org/2000/svg">
```
streetcoder123: Same here

src/client/src/containers/Footer/index.tsx
```diff
@@ -105,15 +107,16 @@ class Footer extends React.Component<Props> {
...
+      intl
     } = this.props;
     e.preventDefault();
     // @ts-ignore
     vscode.postMessage({
       module: EXTENSION_MODULES.GENERATE,
       command: EXTENSION_COMMANDS.GENERATE,
       track: false,
-      text: "Sending generation info...",
+      text: intl.formatMessage(PAYLOAD_MESSAGES_TEXT.SENT_GENERATION_INFO_TEXT),
```
streetcoder123: I am not sure how would this affect but to be consistent with all the other strings, should we have the text as part of the `defineMessages`?

annielovesjs:
> The reset pages icon is not visible on VSCode light mode. I think you need to add VSCode variable css property for its color :)
> 
> image

Nice catch, I added another classname to it and it is the same color as all the other icon on light mode now :)

src/client/src/assets/i-close.svg
annielovesjs: yup - good catch! I forgot to add it in since I didn't use these but will need it for another pr

src/client/src/containers/Footer/index.tsx
jsondoo: Just wondering, should we be localizing strings in the payloads we send to the extension? Or is this a message that's displayed in the client?

annielovesjs: yeah I realized that too - I just changed it to not be localized

src/client/src/containers/RightSidebar/index.tsx
```diff
@@ -196,7 +212,11 @@ class RightSidebar extends React.Component<Props, IRightSidebarState> {
                   optionsData={backendOptions}
                 />
                 <div className={styles.sortablePages}>
-                  {showPages && <SortablePageList />}
+                  {showPages && (
+                    <SortablePageList
+                      handleResetPages={this.resetAllPages.bind(this)}
```
DIGarnier: If you transform the method `resetAllPages` into an arrow function you won't need to bind this explicitly 😁

src/client/src/containers/SortablePageList/index.tsx
```diff
@@ -120,18 +120,26 @@ const SortablePageList = (props: Props) => {
...
+            <button
+              className={styles.resetButton}
+              onClick={props.handleResetPages}
+            >
+              <ResetIcon className={styles.viewIcon} />
+            </button>
+            <button
+              className={styles.hideOrShow}
+              onClick={() => {
+                setMinimized(isMinimized ? false : true);
```
DIGarnier: You could use `!isMinimized`

src/extension/src/utils/vscodeUI.ts
```diff
@@ -30,7 +31,10 @@ export class VSCodeUI extends WizardServant {
   }
 
   async promptUsersToResetPages(message: any) {
-    if (message.payload.pagesLength > 0) {
+    if (
+      message.payload.pagesLength > 0 &&
+      message.text == PAYLOAD_MESSAGES_TEXT.SWITCH_FRAMEWORKS_TEXT
```
DIGarnier: [It's best to always use === to check for value equality](https://www.impressivewebs.com/why-use-triple-equals-javascipt/)

src/client/src/containers/RightSidebar/index.tsx
annielovesjs: LOL thanks mate

streetcoder123 (dismissed):

streetcoder123 (approved):

jsondoo (approved):

SahilTara (approved):

src/client/src/assets/i-hamburger.svg
```diff
@@ -0,0 +1,5 @@
+<svg alt="expand menu" width="20" height="16" viewBox="0 0 20 16" fill="none" xmlns="http://www.w3.org/2000/svg">
+<path d="M20 0H0V0.938889H20V0Z" fill="white"/>
+<path d="M20 7.50391H0V8.44279H20V7.50391Z" fill="white"/>
+<path d="M20 14.5391H0V15.478H20V14.5391Z" fill="white"/>
+</svg>
```
piotrmark, line 5: Is this icon used somewhere? I don't see a reference. In my sidebar PR I kinda hardcoded hamburger without svg icon, so this could be used instead ;)

piotrmark (approved):

PR #940, opened 2019-07-19
Added Character Limit on Page Names

ali-kha:
Using code very similar to that used on the project name limit, added name limits on pages. Now it will stop the user at 50 chars and an error will pop up under the input stating that they cannon pass 50 char. The char limit is set in constants. A lot of whitespace was auto-formatted when the linter ran, the files I touched had never got through linting before.


src/client/src/components/DraggableSidebarItem/index.tsx
```diff
@@ -116,40 +134,55 @@ const DraggableSidebarItem = ({
...
+                    value={text}
+                    disabled={true}
+                  />
+                )}
             </div>
           </div>
+          {pageNameMaxLength && (<div
+            className={classnames({
+              [styles.errorTextContainer]:
+                withIndent || reorderSvgUrl || true,
```
piotrmark, line 162: Always true?

jsondoo: Seems like this code was first checked in here #509 and copied over, but ya prob should take a look at this

jsondoo:
looks good, just error still stays even if u switch pages clicking next or back

src/client/src/components/DraggableSidebarItem/index.tsx
```diff
@@ -116,40 +134,55 @@ const DraggableSidebarItem = ({
                   className={classnames(styles.input, {
                     [styles.azureFunctionNameInput]: isAzureFunction
                   })}
+                  maxLength={maxInputLength === undefined ? undefined : maxInputLength}
```
DIGarnier: This will have the same behavior than using `maxInputLength ` directly

src/client/src/containers/SortablePageList/index.tsx
```diff
@@ -118,7 +116,7 @@ const SortablePageList = (props: Props) => {
           <div className={styles.dropdownTitle}>
             {`${props.intl!.formatMessage(messages.pages)} (${
               pages.length >= 0 ? pages.length : ""
-            })`}
+              })`}
```
DIGarnier: whitespace change. Looks like the person responsible for this branch doesn't have the same automatic linter/formatter installed than the rest of the team

src/client/src/containers/SortablePageList/index.tsx
```diff
@@ -129,15 +127,16 @@ const SortablePageList = (props: Props) => {
             {isMinimized ? (
               <ShowIcon className={styles.viewIcon} />
             ) : (
-              <HideIcon className={styles.viewIcon} />
-            )}
+                <HideIcon className={styles.viewIcon} />
```
DIGarnier: same as ^

src/client/src/components/DraggableSidebarItem/index.tsx
jsondoo: Hello david, nice to see you here LOL

src/client/src/components/DraggableSidebarItem/index.tsx
```diff
@@ -73,6 +80,17 @@ const DraggableSidebarItem = ({
   const handleCloseOnClick = () => {
     idx && handleCloseClick && handleCloseClick(idx - 1); // correction for idx + 1 to prevent 0th falsey behaviour
   };
+
+  const handleKeyDownInput = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    const inputKeyCheck = /^[A-Za-z0-9_\- ]$/
+    let event = e.target as HTMLInputElement;
```
jsondoo: NIT: `const`

src/client/src/components/DraggableSidebarItem/index.tsx
DIGarnier: hola amigo  :smile:

src/client/src/components/DraggableSidebarItem/index.tsx
```diff
@@ -73,6 +80,17 @@ const DraggableSidebarItem = ({
...
+    if (event.value.length === maxInputLength && inputKeyCheck.test(e.key)) {
```
xWiiLLz, line 87: I feel like this approach with onKeyDown is never going to handle all cases.

For example, pressing the CTRL key will not pass your regex, thus removing the max length error prompt 😕
Also, what happens if someone copy-pastes something that has more than `maxInputLength`? I think it would still pass without showing the error.

Another solution would be to make the verification before even calling the `handleChange` like this:
<pre>
const validateInputLength = (
    index: number,
    inputHandler?: (e: any, idx: number) => void
  ) => (event: React.ChangeEvent<HTMLInputElement>) => {
    const { value } = event.target as HTMLInputElement;
    if (maxInputLength && value.length > maxInputLength) {
      return event.preventDefault();
    }

    setPageNameMaxLength(value.length === maxInputLength);
    inputHandler && inputHandler(value, index - 1);
  };
</pre>

You could then refactor the input element from this (lines 139-144):
<pre>
onChange={e => {
  if (handleInputChange && idx) {
    handleInputChange(e.target.value, idx - 1);
  }
}}
onKeyDown={handleKeyDownInput}
</pre>

to this:
<pre>
onChange={validateInputLength(idx, handleInputChange)}
</pre>

src/client/src/components/DraggableSidebarItem/index.tsx
ali-kha: huh... idk what I was thinking there but you are correct

src/client/src/components/DraggableSidebarItem/index.tsx
ali-kha: you see, you would think this would work, and  so did I.... but turns out that `onChange` only works when the input box actually changes. So if I am at the max input box and i hit a key, it doesn't register a change and wont run the code. Also I the CTRL key issue is by design, because its not an input character, you are not trying to input a new character when pressing ctrl so the error shouldn't come up. 

I tried to use the onChange function to begin with, but it wouldn't update unfortunately. The solution in that case is to allow max chars+1 and when they enter that extra key show the error, but that feels less responsive.

src/client/src/components/DraggableSidebarItem/index.tsx
ali-kha: it's not a truth conditional, its an existence one. `reorderSvgUrl` is a string so if it is undefined, then the style is true, else it will inherit the value of reorderSvgUrl. `true` is a default value in a sense

streetcoder123 (approved):

jsondoo (approved):

annielovesjs (approved):

src/client/src/components/DraggableSidebarItem/index.tsx
xWiiLLz: Not sure I understand your point. Have you tried the proposed solution? I tested it and it works better than what was there, handling all cases (even the CTRL & other copy-paste methods).

PR #943, opened 2019-07-19
Fix Quick Start Alignment

streetcoder123:
**Changes**
- Slight fix on the alignment of quick start wand with the text beside it.

image


jcoc611-microsoft (approved):

jsondoo (approved):

jsondoo:
looks way better than before
NIT maybe we can separate the space between wand and text, and move the wand slightly lower

annielovesjs (approved):

PR #946, opened 2019-07-22
Display progress on generation modal

piotrmark:
Changes:
- Disable the button when generation in progress
- Display generation progress.

The generation status message is being passed from coreTS to the extension and then from the extension to the client. Both extension and client just listen to the message and pass it, the message itself is generated in coreTS. Because of that there were 3 ways to do it:
1. Hack the progress on the client side (super hacky)
2. Hack the progress on the extension side and pass in the message to the client (less hacky, but not ideal)
3. Generate progress info in CoreTS and pass it to the extension (best way IMO)

Now, I choose option 2, because it's hackiness is acceptable IMO and option 3 would require changing the CoreTS API contract and probably coordination with WinTS. Instead I intercept the message in the extension and concatenate the progress info based on the Generate message payload.

Take a look and decide whether you want to merge this version or do it properly in CoreTS :)

How to test:
- Generate project with pages and/or Azure services
- Generation modal should look like this:

image

### DIFF
--- a/src/extension/src/signalr-api-module/generateCommand.ts
+++ b/src/extension/src/signalr-api-module/generateCommand.ts
@@ -17,7 +17,8 @@ export class GenerateCommand extends CoreTemplateStudioApiCommand {
 
     connection.on(CONSTANTS.API.GEN_LIVE_MESSAGE_TRIGGER_NAME, genMessage => {
       generatedItemsCount++;
-      const messageWithProgress = `(${generatedItemsCount}/${itemsToGenerateCount}) ${genMessage}`;
+      const percentage = (generatedItemsCount / itemsToGenerateCount) * 100;
+      const messageWithProgress = `(${percentage.toFixed(0)} %) ${genMessage}`;
       this.commandPayload.liveMessageHandler(messageWithProgress);
     });
 
--- a/src/extension/src/signalr-api-module/generateCommand.ts
+++ b/src/extension/src/signalr-api-module/generateCommand.ts
@@ -18,7 +18,7 @@ export class GenerateCommand extends CoreTemplateStudioApiCommand {
     connection.on(CONSTANTS.API.GEN_LIVE_MESSAGE_TRIGGER_NAME, genMessage => {
       generatedItemsCount++;
       const percentage = (generatedItemsCount / itemsToGenerateCount) * 100;
-      const messageWithProgress = `(${percentage.toFixed(0)} %) ${genMessage}`;
+      const messageWithProgress = `(${percentage.toFixed(0)}%) ${genMessage}`;
       this.commandPayload.liveMessageHandler(messageWithProgress);
     });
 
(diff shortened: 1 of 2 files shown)


SahilTara:
Is the goal here to expose the "(i/n) Generating Some Page..."? 

If needed in Core, https://github.com/microsoft/CoreTemplateStudio/blob/dev/code/src/CoreTemplateStudio/CoreTemplateStudio.Core/Gen/GenController.cs this file would be the one to change I'm pretty sure.  Just need to get the length and keep a counter or something.

piotrmark:
> Is the goal here to expose the "(i/n) Generating Some Page..."?
> 
> If needed in Core, https://github.com/microsoft/CoreTemplateStudio/blob/dev/code/src/CoreTemplateStudio/CoreTemplateStudio.Core/Gen/GenController.cs this file would be the one to change I'm pretty sure. Just need to get the length and keep a counter or something.

Yeah, that's why I think it would be best to do it in CoreTS, but I guess this fearure is WebTS-specific and not needed in WinTS

Tanya0609:
I was trying to generate 1 page and the progress is showing i/5. That feels like I am generating 5 pages instead of 1.  (It seems like first 4 are used by generating project. )

If this is the case, will show the percentage an alternative?

piotrmark:
> I was trying to generate 1 page and the progress is showing i/5. That feels like I am generating 5 pages instead of 1. (It seems like first 4 are used by generating project. )
> 
> If this is the case, will show the percentage an alternative?

Yes, it's exactly the case.
I like the idea of percentage and it's a simple change :)
@xtinah-w, what do you think?

xtinah-w:
@piotrmark Yes, I like the percentage as well!

piotrmark:
image

Percentage it is :)

src/extension/src/signalr-api-module/generateCommand.ts
```diff
@@ -17,7 +17,8 @@ export class GenerateCommand extends CoreTemplateStudioApiCommand {
 
     connection.on(CONSTANTS.API.GEN_LIVE_MESSAGE_TRIGGER_NAME, genMessage => {
       generatedItemsCount++;
-      const messageWithProgress = `(${generatedItemsCount}/${itemsToGenerateCount}) ${genMessage}`;
+      const percentage = (generatedItemsCount / itemsToGenerateCount) * 100;
+      const messageWithProgress = `(${percentage.toFixed(0)} %) ${genMessage}`;
```
tejbirsingh: NIT: you should remove the space before the `%` sign.
I've never seen it used that way.
i.e. it should be `25%` and not `25 %`

piotrmark: Oops, thanks! :)

piotrmark: Fixed

SahilTara (approved):

streetcoder123:
> image
> 
> Percentage it is :)

NIT: The space between `25` and `%` looks unusual. Should we get rid of it?

piotrmark:
> > image
> > Percentage it is :)
> 
> NIT: The space between `25` and `%` looks unusual. Should we get rid of it?

Yes, it was fixed in the last commit, sorry for not updating screenshot :)

streetcoder123 (approved):

jsondoo (approved):
looks good, thanks piotr :)

annielovesjs (approved):

PR #949, opened 2019-07-23
Azure forms sequential navigation

piotrmark:
Before:
- All fields enabled when Subscription selected
After:
- Fields enabled sequentially when previous field is not empty

How to test:
- Open all Azure services forms

image

### DIFF
--- a/src/client/src/containers/AppServiceModal/index.tsx
+++ b/src/client/src/containers/AppServiceModal/index.tsx
@@ -507,7 +507,8 @@ const AppServiceModal = (props: Props) => {
             styles.selectionContainer,
             {
               [styles.selectionContainerDisabled]:
-                appServiceFormData.subscription.value === ""
+                appServiceFormData.chooseExistingRadioButtonSelected &&
+                !appServiceFormData[FORM_CONSTANTS.RESOURCE_GROUP.value].value
             }
           )}
         >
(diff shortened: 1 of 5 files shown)


piotrmark:
BTW, the Azure forms code would need some love and refactoring ;)

jcoc611-microsoft (approved):

SahilTara (approved):

streetcoder123 (commented):
@piotrmark Both `Location` and `Number of Functions` become activated once the `App Name` is selected. Is it desired or has it been missed out?

Update: `Number of Functions` is actually always enabled

piotrmark:
> @piotrmark Both `Location` and `Number of Functions` become activated once the `App Name` is selected. Is it desired or has it been missed out?
> 
> Update: `Number of Functions` is actually always enabled

Yes, it was always enabled before, so I left it always enabled.
However I'm not sure if this this intended behavior. @xtinah-w?

jsondoo (approved):

jsondoo:
From what I remember number of functions was always enabled from the start so it was left untouched but it probably makes more sense to disable it?

xtinah-w:
Yeah, it would be great if we can disable it! This was a UX decision made from the previous cohort.

xtinah-w:
Thank you, Piotr, for doing this! We just got feedback from Redmond for the Azure forms so we actually have to change the UX for this now :'( 

We just want to disable the entire form until they fill out subscription because we'll be pre-filling the project name for them and it's weird to have a pre-filled name when the fields are currently disabled.

Thank you for the work!!

jsondoo:
Closing this PR - after our discussion, we will be prefilling some fields so disabling won't make as much sense anymore

Thank you @piotrmark for your beautiful work :)

PR #951, opened 2019-07-23
Add no cloud services summary info

piotrmark:
Changes:
- When no Azure service is selected, sidebar on summary page will display button linking back to Azure page
How to test:
- Go to summary page with and without services selected

image


jcoc611-microsoft (dismissed):

src/client/src/containers/RightSidebar/index.tsx
```diff
@@ -73,6 +74,13 @@ type Props = IRightSidebarProps &
   IDispatchProps &
   InjectedIntlProps;
 
+const hasAzureServices = (services: any) => {
+  for (var key in services) {
```
SahilTara: This PR has made me realize it would be much simpler with getServicesSelector from the wizardSelectionSelector file as this would have been a simple .length > 0. Although the structure of the data is slightly different... So I'll look into it sometime.
Anyways, most nitpicky thing I can say is use const instead of var.

Could maybe even do
```ts
 return Object.values(services).some(service => !_.isEmpty(service.selection));
```
As the body of the function or something of the sort, but its good as it is now too.

piotrmark: Yes! It's not good, var is bad, I'll fix it :D

SahilTara (approved):

streetcoder123 (commented):
There are quite a bit of changes in the en.json and whitelist_en.json files. Just wondering if you ran `npm run manage:translations`?

xtinah-w:
This looks great!

piotrmark:
> There are quite a bit of changes in the en.json and whitelist_en.json files. Just wondering if you ran `npm run manage:translations`?

Yes and that's how these changes were generated :<
Hence the commit message :P 
image
How do you usually solve this?

jsondoo:
> > There are quite a bit of changes in the en.json and whitelist_en.json files. Just wondering if you ran `npm run manage:translations`?
> 
> Yes and that's how these changes were generated :<
> Hence the commit message :P
> image
> How do you usually solve this?

We sometimes manually add strings to the translations file lol

streetcoder123 (approved):

jsondoo (approved):
looks good :)

annielovesjs (approved):
nice!

PR #952, opened 2019-07-24
List page icon was missing in local dev

### LINKED ISSUE #922: List page icon is missing on the pages page and right side bar
To reproduce, on the dev branch cd into src/client and do a `yarn start`
On the pages page, notice that the list page does not have an icon
image

On the right sidebar, notice that the list page is missing the icon:
image

tejbirsingh:
Before:
image

After:
image

This closes #922

### DIFF
--- a/src/client/src/mockData/mockVsCodeApi.ts
+++ b/src/client/src/mockData/mockVsCodeApi.ts
@@ -272,7 +272,7 @@ const mockVsCodeApi = () => ({
                     requiredVisualStudioWorkloads: []
                   },
                   {
-                    emplateId: "wts.Page.React.List",
+                    templateId: "wts.Page.React.List",
                     name: "List",
                     defaultName: "List",
                     description: "Add and remove text from an adaptive list.",


tejbirsingh:
I'm not sure the correct way to link to issues in github like Azure DevOps so I just referenced it #922

SahilTara:
You just need to reference really to link. You can even write "closes #<issue number>" to close the issue when this is merged. Do you know whats up with those plus and minus icons though they seem to be running away from the circle.

tejbirsingh:
I didn't even notice the + and - signs 😲. I can take a look but I would say we should create another issue for it. Then this one keeps a small scope of just the icon.

src/client/src/mockData/mockVsCodeApi.ts
```diff
@@ -272,7 +272,7 @@ const mockVsCodeApi = () => ({
                     requiredVisualStudioWorkloads: []
                   },
                   {
-                    emplateId: "wts.Page.React.List",
+                    templateId: "wts.Page.React.List",
```
piotrmark, line 275: ❤

piotrmark (approved):

piotrmark:
> I didn't even notice the + and - signs 😲. I can take a look but I would say we should create another issue for it. Then this one keeps a small scope of just the icon.

It looks good on my machine :D

image

jcoc611-microsoft (approved):

SahilTara (approved):

jasmingoh (approved):

Issue #897, opened 2019-07-12
"Working" looks like it's an active button when it's generating the project 


xtinah-w:
Need to come up with a new design to respect the blue colour attribute throughout the system
![workingScreenshot](https://user-images.githubusercontent.com/24245311/61093656-6ac8a380-a400-11e9-8b4e-97770203014d.PNG)

SahilTara:
Maybe use the black button here + disable?

xtinah-w:
Thanks for the suggestion, Sahil! I'll look into this and propose it to @jsondoo as an option.

SahilTara:
Think this is closed as of #946

PR #953, opened 2019-07-24
Making the Working button not look like an active button

tejbirsingh:
#897 
When the post-generation modal was working on generating the project, the Working button looked like an active button. I updated the styling to follow @SahilTara suggestion and make it a dark inactive button. If the Template is not failed or generated then it will have dark button styling.

Before:
image

After:
image


tejbirsingh:
Closing as @piotrmark has a better change for this in #946

PR #954, opened 2019-07-24
Make sidebar hide and show

piotrmark:
Changes:
- Remove unused state from sidebar
- Add state to control sidebar open/hidden

How to test:
- Click arround. Sidebar should be hidden by default on frameworks page and not visible on welcome and details pages. You cannot hide sidebar on summary page.

sidebar


src/client/src/appStyles.module.css
```diff
@@ -14,10 +14,12 @@
   height: calc(100vh - 124px);
   justify-content: flex-start;
   position: relative;
+  padding-right: 500px;
 }
 
 .centerViewAzurePage {
   padding: 0;
+  padding-right: 500px;
```
tejbirsingh: Will this be responsive?

src/client/src/containers/RightSidebar/index.tsx
```diff
@@ -64,8 +66,10 @@ interface IRightSidebarProps {
 }
 
 interface IRightSidebarState {
-  frontendDropdownValue?: ISelected | undefined;
-  items: string[];
+  sideBarOpen: boolean;
+  // Since componentWillReceiveProps lifecycle method is now depricated and we don't have acces to prevProps in getDerivedStateFromProps,
```
tejbirsingh: NIT: type `deprecated`

tejbirsingh: It arguably should be part of the state and not props anyways since it is only part of the rightsidebar

src/client/src/containers/RightSidebar/index.tsx
```diff
@@ -151,82 +188,104 @@ class RightSidebar extends React.Component<Props, IRightSidebarState> {
     } = this.props;
     const { formatMessage } = intl;
     const { frontendOptions, backendOptions } = contentOptions;
+    const { sideBarOpen } = this.state;
+
+    if (pathname === ROUTES.PAGE_DETAILS || pathname === ROUTES.NEW_PROJECT) {
```
tejbirsingh: I feel like you could use `pathname === ROUTES.PAGE_DETAILS || pathname === ROUTES.NEW_PROJECT` somewhere else to set `sideBarOpen`. Because you are using that logic (`pathname === ROUTES.PAGE_DETAILS || pathname === ROUTES.NEW_PROJECT` ) to also display or not display the sidebar. Then this conditional could just be `if(sideBarOpen) return null;

tejbirsingh:
I don't know if it's a good experience to have a different default hide/show the sidebar on different pages. I feel like it should be consistent and then let the user decide what they want by clicking to open or close it. 

Alternatively if we keep it this way, maybe you should auto-open the sidebar once the user selects one framework. So it's closed by default until they select something.
What do you think?

src/client/src/containers/RightSidebar/index.tsx
piotrmark: This comment is for the locationPathname string. I will make it clearer :)

src/client/src/containers/RightSidebar/index.tsx
piotrmark: Actually the logic is quite different here.
When !sideBarOpen, we display the hamburger, while on details and new_project we don't render anything.

piotrmark:
> I don't know if it's a good experience to have a different default hide/show the sidebar on different pages. I feel like it should be consistent and then let the user decide what they want by clicking to open or close it.
> 
> Alternatively if we keep it this way, maybe you should auto-open the sidebar once the user selects one framework. So it's closed by default until they select something.
> What do you think?

I think it's fine, but since I'm not UX expert, I'll defer to @xtinah-w :)

xtinah-w:
@tejbirsingh Yeah, that's a really good point. I thought about it before and maybe we can just experiment with always having it hidden (on the frameworks, pages, and azure page) and always have it show on summary. 

Originally, I didn't want to hide it since I wanted to make it obvious that you can interact with the right side bar, but hiding it may actually be a better indicator that there are other things you can do with it now (especially with the addition of the + and - button on the "Add Pages" page).

In this case, the only time it will show up on Pages & Azure would be when people start adding stuff, since we also got feedback from Redmond to hide sections not directly applied to their projects (i.e. right sidebar would hide more things apart from the actual section needed to be shown). 

@piotrmark let me know your feedback on this! If it takes too long to hide it by default on the other 2 pages, that's fine too!

piotrmark:
@xtinah-w Sure, it will be a quick change :)
One thing I'd like to clarify: should the sidebar open automatically when user adds a page or Azure service?

SahilTara:
> I don't know if it's a good experience to have a different default hide/show the sidebar on different pages. I feel like it should be consistent and then let the user decide what they want by clicking to open or close it.
> 
> Alternatively if we keep it this way, maybe you should auto-open the sidebar once the user selects one framework. So it's closed by default until they select something.
> What do you think?

I agree, I think we should have it be closed by default but once you start adding things it auto opens. When I go to lets say the azure page after adding a page or two I wouldn't want to have to click back to delete a page or reopen the menu that was already open, just to remove a page. Same applies with azure. I think its more reasonable to keep it open once the user adds stuff. @xtinah-w

Also think we might want to open a new issue regarding the placement of things since the hiding of the right sidebar results in some weird locations for things like sign out.

src/client/src/appStyles.module.css
piotrmark: This is the width of the sidebar. I had to add it here as I changed it's position to fixed.
It keeps the responsiveness of main content :)

src/client/src/containers/RightSidebar/index.tsx
tejbirsingh: What you're saying makes sense but it's still kind of confusing. There are three states of the side bar. I guess I find it confusing to use `sidebarOpen` but there's a third state in there.

tejbirsingh: I feel like that "should we render it at all" logic should be passed in as a prop `disabled` and then you do the pathname=== logic in the parent.

piotrmark: I agree the code may be confusing.
The problem here is that the parent is App component and I don't want to put this logic there.
I'll try with disabled state, but calculated in Sidebar component.

piotrmark:
New version:
Sidebar is hidden by default and opens automatically when user adds a page or a service.
When user manually closes or opens the sidebar, it enters 'user-controled' state and from this point is never automatically opened or closed.

Of course, it stays always open on summary page.

src/client/src/containers/RightSidebar/index.tsx
piotrmark: After recent changes, it doesn't exist anymore :)

src/client/src/containers/RightSidebar/index.tsx
piotrmark: At the second thought, I don't see why not put this logic in App component, we already check pathname there for splash background :)

jsondoo:
Right side bar works well, just two small things

image
The right side bar feels a bit bigger on all of the pages to me

image
Scrollbar also seems to be hidden on certain steps 
(we can take these tasks in a separate PR since you probably don't have capacity haha)

xtinah-w:
Responsive screen states design - ideally everything should scale! 

I've also moved the sign-in component to the left and added a background colour (same as the cards) to give it some separation. Please let me know if anyone has any questions or concerns! :) 

Big-responsiveSidebar-closed
Big-responsiveSidebar-open
small-responsiveSidebar-closed
small-responsiveSidebar-open
small-responsiveSidebar-loggedIn

src/client/src/containers/RightSidebar/index.tsx
```diff
@@ -34,10 +34,12 @@ import {
   ROUTES,
   EXTENSION_COMMANDS,
   EXTENSION_MODULES,
-  PAYLOAD_MESSAGES_TEXT,
+  KEY_EVENTS,
+  PAYLOAD_MESSAGES_TEXT
```
SahilTara: You're missing a , here.

annielovesjs:
will reopen once i start making the changes next week

src/client/src/containers/RightSidebar/index.tsx
annielovesjs: added!

annielovesjs:
changes (should match the design now):
1) on smaller screens, right sidebar overlaps all the elements and expand to fill the page
2) on larger screens, right sidebar squishes everything to the left to make room for itself
3) added smoother transition to opening the sidebar
4) remove/added additional padding

Tanya0609 (commented):
Great Work!! 👍 Right side bar is responsive for all pages. Just a small problem with **sign out**. It has been cut off for some cases.

1. when you are in regular screen size, sign in. When it success, the sign out is cut off
Screen Shot 2019-08-07 at 9 29 16 PM
After that, when you open the right side bar, the sign out will turn to okay.
Screen Shot 2019-08-07 at 9 30 09 PM
2. the text space is wired on mac when shrink to smaller size. ( I think this should be a different PR that fix all the inconsistency fonts on Mac)
Screen Shot 2019-08-07 at 9 26 45 PM

annielovesjs:
> Great Work!! 👍 Right side bar is responsive for all pages. Just a small problem with **sign out**. It has been cut off for some cases.
> 
> 1. when you are in regular screen size, sign in. When it success, the sign out is cut off
> 
> Screen Shot 2019-08-07 at 9 29 16 PM
> 
> After that, when you open the right side bar, the sign out will turn to okay.
> Screen Shot 2019-08-07 at 9 30 09 PM
> 
> 2. the text space is wired on mac when shrink to smaller size. ( I think this should be a different PR that fix all the inconsistency fonts on Mac)
> Screen Shot 2019-08-07 at 9 26 45 PM

Thank you :) and also good catch! I just fixed the overlapping signout issue

Tanya0609 (commented):
is there a reason why CoreTemplateStudio is part of the change?

ali-kha:
On the pages page, if you add enough pages to make the right side bar scroll-able, the border disappears.
file

ali-kha:
The view lisnaces button on the summary page doesn't work, looks like the SVG is overlapping.

Tanya0609 (changes requested):
Could you remove the changes to CoreTemplateStudio files?

If you rebase to the newest dev branch and do **git submodule update**, then those files won't show up as changes. Hope this helps.

annielovesjs:
@ali-kha  I can't seem to reproduce this error (disapearing border) - can you build and try again?

annielovesjs:
@ali-kha the overlap should be fixed now

src/client/src/containers/RightSidebar/index.tsx
```diff
@@ -104,6 +106,31 @@ class RightSidebar extends React.Component<Props, IRightSidebarState> {
     selectWebApp: () => {},
     selectPages: () => {}
   };
+
+  state: IRightSidebarState = {
+    isSidebarOpen: false,
+    isSidebarUserControlled: false
+  };
+
+  static getDerivedStateFromProps(
```
Tanya0609, line 115: Does this function use anywhere?

annielovesjs: I think this was initially added in case we wanted to open up the right side bar on certain pages (ex. azure page and pages page) instead of defaulting it to close

annielovesjs: Currently it is not being used, but we can make the right side bar display in another pr which would use this function

jsondoo: isn't this a lifecycle method? or is this method not being called at all?

jsondoo: @piotrmark do you want to comment?

annielovesjs: never mind it's called whenever the component is rendered

piotrmark: Yes, It's a lifecycle method, rather important one :D
It opens sidebar when user adds a page or Azure service

jasmingoh:
Small unexpected behavior:
seems like the cloud database section moves when the sidebar shows/hides

annielovesjs:
> Small unexpected behavior:
> seems like the cloud database section moves when the sidebar shows/hides
> 

@jasmingoh  fixed!

annielovesjs (approved):

jasmingoh (approved):

ali-kha (approved):

jsondoo (approved):

Tanya0609 (approved):

PR #956, opened 2019-07-25
AppService and AzureFunctions modal were not correctly setting inputs as 'disabled'

### LINKED ISSUE #947: You can type into disabled field
Azure services forms - when the input text field is disabled, we can still type into it:
image

tejbirsingh:
The comparison used to be an object to string match. This fixed #947 

Before:
image

After:
image


jcoc611-microsoft (approved):
Nice!

SahilTara (approved):

piotrmark (approved):

jsondoo (approved):

PR #960, opened 2019-07-29
Don't reroute to pages after changing frontend framework

annielovesjs:
*this will also work the same for the reset button in the other pr since they share the same extension command*
Changes/how to test
- select some pages and then change the frontend framework -  you will remain on the same page


annielovesjs:
will be reviewed after the pr for saving pages on frontend change goes in

PR #966, opened 2019-07-29
Azure Functions: when there is 1 function left, user should not be able to remove it 

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/30870

**Changes**
- Try to add 2-3 Azure functions.
- Then from the right side bar, remove each function 1 by 1. The last function remaining would not have the `X` sign next to it.

**With 3 Functions**
image

**With 1 function remaining**
image


jcoc611-microsoft:
Just wondering...if you add some Azure functions and then decide you don't want any, how do you go about deleting all of them?

streetcoder123:
> Just wondering...if you add some Azure functions and then decide you don't want any, how do you go about deleting all of them?

We can do that by clicking the `X' button on the side of functions name (qadsfd) in the screenshot :)

jasmingoh (approved):
works!

piotrmark (approved):

SahilTara (approved):

jsondoo (approved):

PR #969, opened 2019-07-30
Accessibility issues - images (alt, roles, titles)

annielovesjs:
**Addressed these accessibility tasks:**
- Provide descriptive alt tags for all images 
- Ensure that img elements with an svg source includes the role="img"attribute. 
- Better tooltip text/title

**What you can test**
Hover over all the following svg/imgs:
- for right side bar on pages page eye icon (Before: i-show, i-hide, After: show options, hide options)
- for right side bar on pages refresh icon (Before: null, After: reset pages)
- home page image (Before: homesplash8, After: An octowizard creating his full-stack project in space)
- dependency checker (Before: green check mark, After: null)
- summary page image (Before: summarysplash, After: An octowizard launching his project in space)
- home page quickstart wand (Before: i-wand, After: Quick start your webapp)


jcoc611-microsoft (dismissed):

src/client/src/assets/i-reset.svg
```diff
@@ -1,3 +1,3 @@
-<svg alt="reset all pages" width="12" height="13" viewBox="0 0 12 13" fill="none" xmlns="http://www.w3.org/2000/svg">
+<svg alt="reset all pages" width="12" height="13" viewBox="0 0 12 13" fill="none" xmlns="http://www.w3.org/2000/svg"><title>Reset pages</title>
```
SahilTara: This one doesn't have a role is this intended?

annielovesjs: Oops, I accidentally missed that but I just added the role in :)

streetcoder123 (dismissed):

piotrmark (commented):
Is it OK to have English alt text in localized versions of the app?

piotrmark (dismissed):

SahilTara:
@piotrmark For some cases like "web template studio" or something it would be okay I think, but for other things it may not be. Like if im using webts with a diff language it would be nice if the hovers were in that language too. Also side note will the interns localize before they leave at end of august? If so should internationalize the alt tags and anything else

SahilTara (dismissed):

annielovesjs:
> @piotrmark For some cases like "web template studio" or something it would be okay I think, but for other things it may not be. Like if im using webts with a diff language it would be nice if the hovers were in that language too. Also side note will the interns localize before they leave at end of august? If so should internationalize the alt tags and anything else

hmm, yeah I could localize the alt tags by wrapping the svgs in a div and assigning the localized alt to that div; however for the tooltip/hover text, I've tried a couple methods to append the title outside of svg file so it can be localized (I've tried svgr title props, adding <title> elements inside the svg tag) but it doesn't seem to render. I would love to know if you have ideas on how to approach this! The only thing that I can think of that will for sure work is directly including the svg in the ts file so I can directly change the title attribute instead of importing the svg as a react component

streetcoder123 (dismissed):

piotrmark:
> > @piotrmark For some cases like "web template studio" or something it would be okay I think, but for other things it may not be. Like if im using webts with a diff language it would be nice if the hovers were in that language too. Also side note will the interns localize before they leave at end of august? If so should internationalize the alt tags and anything else
> 
> hmm, yeah I could localize the alt tags by wrapping the svgs in a div and assigning the localized alt to that div; however for the tooltip/hover text, I've tried a couple methods to append the title outside of svg file so it can be localized (I've tried svgr title props, adding <title> elements inside the svg tag) but it doesn't seem to render. I would love to know if you have ideas on how to approach this! The only thing that I can think of that will for sure work is directly including the svg in the ts file so I can directly change the title attribute instead of importing the svg as a react component

Ok, we're not localizing the app now :)

piotrmark (dismissed):

Tanya0609 (changes requested):
These two icons are not consistent. Functions has alt, but Cosmos doesn't. 

Also, the framework icons don't have alt text neither.

alt-text

annielovesjs:
@Tanya0609 just added it in :)

ali-kha (dismissed):

Tanya0609 (dismissed):

SahilTara (dismissed):

piotrmark (dismissed):

jcoc611-microsoft (dismissed):

streetcoder123 (dismissed):

ali-kha (dismissed):

ali-kha (approved):

jasmingoh (approved):

SahilTara (approved):

piotrmark (approved):

PR #970, opened 2019-07-30
App Service - create App Service Plan

jasmingoh:
This PR adds a method for creating an App Service Plan (ASP) which is required for creating a web app.

- The payload from client side looks like this:
image

- A basic tier (B1) linux ASP will be create when App Service is selected
    - Reason for not choosing the free tier (F1) ASP is because only one F1 ASP is allowed per subscription. This means that we will need to handle the logics of deleting and/or reusing the free ASP, and that requires a lot more design work and logics
    - B1 ASP is only free for 30 days - we will include a note in the form to notify our users about the free trial (will be addressed in another PR)
- The extension handles the logic of generating a valid ASP name
    - Refactored and reused the code for generating valid resource group name
- The ASP will be deployed into the same resource group as the App Service (web app)
    - This means that the method that creates the ASP can only be called after all new resource groups are created
___
**Test:**
 
- [x] Insert `await AzureServices.createAppServicePlan(message.subscription);` before the return statement in `sendAppServiceSubscriptionDataToClient` (line 129) in `azureServices.ts`.

- [x] Replace the aspSelection in `createAppServicePlan` (line 360) in `azureServices.ts` with 
```
    const aspSelection: AppServicePlanSelection = {
      subscriptionItem: AzureServices.userAppServiceSubsctiptionItemCache,
      resourceGroup: "jasmin-centralus",
      name: "<your-name-here>"
    };
```
- [x] Launch the extension in preview mode, and click on "Add Resource" for App Service in the Azure page
- [x] Select the "GiV.Hackathon" subscription (clicking on the subscription will create the ASP - **this behavior is only for testing**)
- [x] Go to the Azure portal and see if the ASP is created under resource group **jasmin-centralus**
    - The ASP name will look something like `<project name>_<asp>_<timestamp>`
- [ ] **Make sure to delete the resources you created** :)

___
**Note:** ASP creation may fail if you're deploying into a resource group containing an Azure Function App because our function app is using another type of ASP (kind: functionapp)

To avoid this, we will (1) only all our user to only choose an App Service OR a Function App and (2) recommend our users to always deploy their App Service to a new resource group


src/extension/src/azure/azure-app-service/appServiceProvider.ts
```diff
@@ -75,4 +91,68 @@ export class AppServiceProvider {
...
+  // Creates a Basic Tier App Service Plan (ASP)
+  public async createAppServicePlan(
+    aspSelection: AppServicePlanSelection
+  ): Promise<string> {
+    this.setWebClient(aspSelection.subscriptionItem);
+    if (this.webClient === undefined) {
+      throw new AuthorizationError(CONSTANTS.ERRORS.WEBSITE_CLIENT_NOT_DEFINED);
+    }
+    const appServicePlanSelection: AppServicePlan = {
+      kind: OS.Linux,
```
Tanya0609: Maybe it maybe good if you could put a comment here to indicate we choose the default values or those are default values for future maintenance. You don't need to take this advice LOL

src/extension/src/azure/azure-app-service/appServiceProvider.ts
```diff
@@ -75,4 +91,68 @@ export class AppServiceProvider {
...
+    return !exist;
+  }
+
+  private async checkASPExistence(name: string): Promise<boolean> {
+    if (this.webClient === undefined) {
+      throw new AuthorizationError(CONSTANTS.ERRORS.WEBSITE_CLIENT_NOT_DEFINED);
+    }
+    const allASP: AppServicePlanCollection = await this.webClient.appServicePlans.list();
+    return allASP.some(asp => {
+      return asp.name === name;
```
Tanya0609, line 155: Great work :) 

Just curious, will we need another **return** statement inside this .some?

jasmingoh: thanks :)

Yup, some executes a callback function, so we do need the return

Tanya0609 (approved):

piotrmark (approved):
Amazing job!

jsondoo (approved):

Issue #971, opened 2019-07-30
Users should only be allowed to create one Azure cloud hosting at a time


jsondoo:
Briefly mentioned in the note of #970 

ASP means App Service Plan (https://docs.microsoft.com/en-us/azure/app-service/overview-hosting-plans)

Currently, WebTS creates an ASP (kind: functionapp) when it deploys an Azure Function App. It looks something like this:
![image](https://user-images.githubusercontent.com/20333350/62169293-bb833c00-b2dc-11e9-9163-288c4f8f20c5.png)

The App Service Plan required for creating a web app is different. To create a web app, we need a linux ASP, which looks like this:
![image](https://user-images.githubusercontent.com/20333350/62168871-6d216d80-b2db-11e9-8a23-a5bb9dc63932.png)

The best practice when using Azure services is to group all resources required for a project into the same resource group.

But these 2 kind of ASP cannot exist in the same resource group.
___
We have 2 solutions:
(1) Deploy the Azure Functions to the same ASP as the web app
(2) Only allow users to select either Azure Functions OR App Service

(1) requires a lot more investigation. The development work required will be out of scope for our internship. Also the use case for having Azure Functions and App Service in 1 project is quite rare. Hence, we decided to go with (2), which is the reason for this issue.

xtinah-w:
![oneHostingService](https://user-images.githubusercontent.com/24245311/62167889-09964080-b2d9-11e9-896a-54bc398aa097.png)

jasmingoh:
Refer to the edited description of this issue.

streetcoder123:
@jsondoo This has been fixed, so should we close this?

PR #975, opened 2019-07-30
Allow one azure hosting service at a time

jsondoo:
**Changes**
- Once a hosting service is added, all the other hosting services should be disabled so that only one can be added at a time

**How To Test**
- Adding App Service should disable Azure Functions and vice versa
- Removing a service should enable both cards

Issue #971

image


jasmingoh (approved):
works!

just wondering, should you tell users that they can only select 1 hosting service on the Azure page?

piotrmark (approved):

jsondoo:
> works!
> 
> just wondering, should you tell users that they can only select 1 hosting service on the Azure page?

@jasmingoh yea we should, i will be doing string changes in a different PR

streetcoder123 (approved):

PR #976, opened 2019-07-30
Automate Population of Pages on Framework Change

### LINKED ISSUE #957: Don't reset pages on change in Framework Selection
Currently, if the user changes front-end framework after selecting pages, the pages are reset to just one blank page. The pages for the new framework should auto-populate based on user's currently selected pages.

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/31620

**Changes**
- Changing frameworks after adding a bunch of pages no more resets the pages. Pages for the new framework are automatically populated based on the selection of new front-end framework.

closes #957


src/client/src/containers/RightSidebar/index.tsx
```diff
@@ -123,17 +125,46 @@ class RightSidebar extends React.Component<Props, IRightSidebarState> {
...
-    const { frontendFramework, pages } = this.props.selection;
-    const { vscode } = this.props;
+    const { frontendFramework, backendFramework, pages } = this.props.selection;
+    const {
+      vscode,
+      selectPages,
+      frontEndOptions,
+      selectFrontendFramework
+    } = this.props;
+    console.log(JSON.stringify(option));
```
jasmingoh: should there be console.log here?

src/client/src/containers/SelectFrontEndFramework/index.tsx
```diff
@@ -63,19 +70,32 @@ class SelectFrontEndFramework extends React.Component<Props> {
       selectedFrontendFramework.internalName &&
       selectedFrontendFramework.internalName !== option.internalName
     ) {
+      console.log(JSON.stringify(selectedPages));
+      console.log(option.internalName);
+      console.log(selectedFrontendFramework.internalName);
```
jasmingoh: line 73, 74 and 75: should there be console.log here?

src/client/src/containers/RightSidebar/index.tsx
streetcoder123: Whoops! Thanks for spotting that

src/client/src/containers/RightSidebar/index.tsx
```diff
@@ -123,17 +125,45 @@ class RightSidebar extends React.Component<Props, IRightSidebarState> {
...
+        command: EXTENSION_COMMANDS.GET_PAGES,
         payload: {
-          internalName: option.value,
-          pagesLength: pages.length
+          projectType: WIZARD_CONTENT_INTERNAL_NAMES.FULL_STACK_APP,
+          frontendFramework: option.value,
+          backendFramework: backendFramework.internalName
+        }
+      });
+      const newPages: ISelected[] = pages.map(page => {
```
SahilTara, line 180: not sure if i'm understanding but does this mean if you switch frameworks with [master detail, blank, list] in angular to react you will still have [master detail, blank, list]? If so great UX improvement, but wouldn't this also add in overhead to anyone who wants to contribute a page. Now one must know React, Angular and Vue before they can add a new page to webts. Or the pages must remain synchronized 100% of the time. 

Do you think that this is okay? Assuming that you intended that the team kept pages synchronized anyways, but feels like its harder for someone outside the team to contribute pages.

streetcoder123: Hmm, I see. I had this discussion with the team as well. But, we decided to go ahead with this because this would mean a huge improvement as far as user experience is concerned.
I agree that pages would need to be synchronized across all frameworks now. Yeahh, I am not very sure though.

SahilTara: Yeah this is the main reason why we didn't do this, its a trade off either way. One thing that would be the best of both worlds is to check if it exists before adding it. But I don't know if that additional complexity is wanted?

streetcoder123: That could be a possibility. However, I would keep it as is in this PR. I shall create an issue for that to be done in the next Sprint though. Thanks for the insight - that seems good :)

annielovesjs (dismissed):

Tanya0609 (changes requested):
I randomly went back and forward in the steps, also frequently changed the frameworks choice, and tried rename and reorder the pages, this is what happened

- master detail page somehow be deleted
- there are gaps in the front end 
- when drag pages around, the naming got mess up
test-pages

jsondoo (dismissed):

Tanya0609 (dismissed):

annielovesjs (approved):

jsondoo (approved):

Tanya0609 (approved):

PR #977, opened 2019-07-30
Populate output path when extension is launched

jasmingoh:
Related PBI: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/29106

Currently, users have the option to go into the Settings and choose a default output path. But this feature is very hidden, hence we decided to always have a default output path for our users.

This PR provides a default output path (the home directory) for users when the extension is launched. Users still have an option to overwrite the output path we generate by changing it in the extension settings (`Ctrl + ,`)
___
**Test:**
- [ ] Launch the extension in different OS and make sure that project can be generated successfully without changing the output path
- [ ] Go to the extension setting, fill in the "Change Save To Location", launch the extension and make sure that the output path is overwritten on the home page


src/extension/src/controller.ts
```diff
@@ -190,21 +194,30 @@ export class Controller {
     });
   }
 
+  private static getDefaultOutputPath() {
+    const userOS = os.platform();
```
jcoc611-microsoft: Would `os.homedir();` work better here?

src/extension/src/controller.ts
```diff
@@ -190,21 +194,30 @@ export class Controller {
...
+      }
+    });
+  }
+
+  private static isUnix(userOS: string): boolean {
+    return userOS === "linux" || userOS === "darwin";
+  }
+
   private static loadUserSettings() {
-    let outputPathDefault = vscode.workspace
```
jcoc611-microsoft, line 182: Would it still be valuable to be able to override the default directory on the extension settings?

jcoc611-microsoft (changes requested):
some comments

src/extension/src/controller.ts
jasmingoh: Good questions, I'll talk to the team about it

Update: Talked to the team about it, and we decided to add the setting back so that users can override the output path we generate

src/extension/src/controller.ts
jasmingoh: yup, for sure, I didn't know that method existed

src/extension/src/controller.ts
```diff
@@ -1,4 +1,6 @@
 import * as vscode from "vscode";
+const os = require("os");
```
streetcoder123: Is it intentional that we don't use the `import` syntax instead of `require` syntax to get the `os` module? If not, I think we should use the `import` syntax to be consistent with other imports.

jasmingoh: Yes, it is intentional
https://nodejs.org/api/os.html#os_os

The alternative is this: `import * as os from "os";`

jasmingoh: @piotrmark what are you thoughts on this?

piotrmark: I'd say, let's be consistent.
So `import * as os from "os";`

src/extension/package.json
```diff
@@ -53,10 +53,10 @@
       "type": "object",
       "title": "%webTemplateStudioExtension.commands.wts%",
       "properties": {
-        "wts.defaultOutputPath": {
+        "wts.changeSaveToLocation": {
           "type": "string",
           "default": "",
-          "description": "Load a default output path on wizard launch."
+          "description": "Change your save to location for your project."
```
jsondoo: Confirmed this work on windows :)

We discussed this string with the team and we think `Set default project location for your project` would be more clear because the word `change` might be confusing. What do you think @jasmingoh

streetcoder123 (approved):
Great!

jcoc611-microsoft (approved):

jsondoo (approved):

PR #984, opened 2019-07-31
VueJS: Use short hands for common directives

ali-kha:
This PR is based on expert feedback. Vue has a short cut for the v-on and v-bind directives, being `@` and `:` respectively. Updated generated code to use them instead of full syntax. Also there are minor linting updates. To test generate a vue app with all 4 pages, there should be no `v-on` or `v-bind` directive in the generated code and the ui should work as expected.


SahilTara (approved):
ooo this PR is good to know for the future. Nice job!

streetcoder123 (approved):

jasmingoh (approved):

piotrmark (approved):

PR #992, opened 2019-08-01
VueJS: Use vue absolute paths

ali-kha:
This PR is based on expert feedback. Vue has a short cut for absolute pathing. putting `@` at the beginning of a path will set it to the root directory of the project (where `app.vue` is located in this case)


streetcoder123 (approved):
👍

jasmingoh (approved):

piotrmark (approved):
I like Vue more and more :D

PR #993, opened 2019-08-01
Default to Automation of Resource Group

jsondoo:
**Changes**
- All azure forms default to automation of resource group

**How To Test**
- change that create new resource group for me is selected by default

image


streetcoder123 (commented):
@jsondoo Just out of curiosity, why are we changing this?

jasmingoh:
> @jsondoo Just out of curiosity, why are we changing this?

Mainly because of App Service Plan (ASP). It is required when users choose App Service, but it cannot be created in the same resource group which contains any Function App created by our extension. This is because our Function App runs on another kind of ASP and this causes conflict.

I wanted to include a note to suggest (beginner) Azure users to always choose "Create a new resource group", but @xtinah-w said that it is not obvious enough, and defaulting the resource group option would be better.

Plus, it is best practice to put all Azure resources required for a project in the same resource group :)

jasmingoh (approved):
works!

streetcoder123 (approved):

jsondoo:
@njindal239 Yea, we had a discussion on this, main reason is that we want it to be more beginner friendly and automation of RG does that :)

markAtMicrosoft (approved):

PR #1000, opened 2019-08-01
Default Project Name for the user

### LINKED ISSUE #981: Create a default Project Name
Along with default output path, we should also have a default project name to make the experience even faster.

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/31967

**Changes**
- Added a default Project Name for the user

image

closes #981


src/extension/src/constants.ts
```diff
@@ -268,6 +268,8 @@ export const CONSTANTS = {
   VALIDATION_LIMIT: 3
 };
 
+export const APP_NAME_VALIDATION_LIMIT = 20;
```
jasmingoh: would it is ok if you change this constant to `PROJECT_NAME_VALIDATION_LIMIT` cause app name sounds like azure related stuff lol

streetcoder123: Sure. lol

src/extension/src/controller.ts
```diff
@@ -167,6 +171,31 @@ export class Controller {
     }
   }
 
+  private static getProjectName() {
+    const userOutputPath = vscode.workspace
+      .getConfiguration()
+      .get<string>("wts.changeSaveToLocation");
+    const outputPath: string = userOutputPath ? userOutputPath : os.homedir();
+    const defaultAppName = "myApp";
+    let newAppName = "myApp";
```
SahilTara: why not say newAppName = defaultAppName?

src/extension/src/controller.ts
```diff
@@ -167,6 +171,31 @@ export class Controller {
...
+    let count = 1;
+
+    while (
+      fs.existsSync(path.join(outputPath, newAppName)) &&
+      count <= PROJECT_NAME_VALIDATION_LIMIT
+    ) {
+      newAppName = `${defaultAppName}${count}`;
```
SahilTara, line 187: So the user can only have 20 projects generated for them by default?

Why not make it myApp_SomeRandomStringSeed? It makes it less likely for collisions as well.

src/extension/src/controller.ts
```diff
@@ -167,6 +171,31 @@ export class Controller {
...
+      count++;
+    }
+    if (count > PROJECT_NAME_VALIDATION_LIMIT) return;
```
SahilTara: Nit: This return is hard to notice at first glance. Can you move it to the next line at least? (Don't have to change if you feel the current way is better).

src/extension/src/controller.ts
streetcoder123: yup, makes sense

src/extension/src/controller.ts
streetcoder123: We were thinking about coming up with an app name. We decided to proceed with this pattern since this is how many editors go about giving default names to save files. I was also thinking about the random string appended to app name. The thing with that is it is not very user friendly to see a weird project name being already displayed as default.
For the number of projects, definitely, we can increase that to 50 or maybe 100 but I think that should be a safe choice to stop on.

src/extension/src/controller.ts
streetcoder123: Yeah, I will make it more noticeable :)

src/extension/src/controller.ts
streetcoder123: I have increased the limit to 50 for now.

SahilTara: Sounds good

SahilTara (approved):
Code looks good!

src/extension/src/controller.ts
jcoc611-microsoft: If you're worried about performance, you could do a binary search here. Start at 1, if not exists you can use that, else go to 2, then to 4, then to 8 etc. For the first one you find that doesn't exist, you can move back logarithmically. If it is 8 for example, then the real one is in (4, 8] so you can try 6 first, then 7 and then 8.

streetcoder123: That's a good idea. However, there is a possibility that the user somehow has the following folders: `myApp`, `myApp1` ,`myApp2`, `myApp4`. In this case, we won't be able to recognize that we have `myApp3` available for using as a default project names.

piotrmark: I think it's ok to generete myApp5 in this case :)

jsondoo (approved):
I think this should be a different PR but after clicking `Create New Project` it has no default name (or default output path) otherwise looks good

Tanya0609 (approved):

annielovesjs (approved):

src/extension/src/controller.ts
streetcoder123: Talked with the team, this approach might lead to several problems as far as user experience is concerned as well. For instance, hypothetically, if the user has myApp, myApp1, myApp2, myApp4, myApp8, myApp16 and mypp32, we will create an app name that would be greater than 32 in this case, whereas we might be good to create myApp3. 

It won't take much to do a binary search but the team does not see a value in it. Nevertheless, we can discuss this next week if we are thinking something wrong. We are merging this for now to have default name for the release. :)

PR #1001, opened 2019-08-01
App Service - create new resource group when selected

jasmingoh:
This PR handles the creation of new resource group when user selects "Create a new resource group" in the App Service form.

**Test:**
- Launch the extension in preview mode
- Take note of the project name
- On the Azure page, select App Service and choose "Create a new resource group"
- Create the project
- Go onto Azure portal and make sure that the resource group is created. The name of the newly created resource group should be `<project_name>_<timestamp>`


jasmingoh:
Reopened it here again #1002

PR #1003, opened 2019-08-02
Add compliance links in extension settings

### LINKED ISSUE #998: Extension Settings: Add Privacy Statement and Terms of Use
We will be adding Privacy Statement and Terms of Use to our extension settings since we are an open-source project

jsondoo:
This adds two new links in the vscode extension settings (privacy statement and terms of use)

image

If anyone knows how to not show the `Edit in settings.json` please let me know - all the other types I've tried seem to come with a button or an input field
close #998


src/extension/package.json
```diff
@@ -62,6 +62,10 @@
           "type": "boolean",
           "default": false,
           "description": "Enable upcoming features; Warning: may cause stability issues."
+        },
+        "wts.legalResources": {
+          "type": "null",
```
Tanya0609, line 67: just curious, why the type is null?

src/extension/package.json
```diff
@@ -62,6 +62,10 @@
...
+          "markdownDescription": "[Privacy Link](https://privacy.microsoft.com/en-US/privacystatement) | [Terms of Use](https://www.microsoft.com/en-us/legal/intellectualproperty/copyright/default.aspx)"
```
SahilTara, line 68: This seems like a weird place to put such info. Almost as if you were hiding it haha. Why not put it on the wizard or something?

xtinah-w: This was a decision made by Stephane, since this is an open-source project, people can use our product as they please and the best way to handle this was to put it in settings to have a balance between the nature of the product and always having it there.

src/extension/package.json
jsondoo: So for the type there is `null`, `array`, `number`, `object`, `string`, `boolean`, `object` according to VSCode autocomplete. Some of them show a checkmark like the boolean, some of them show an input field like the number and string. Rest of them show the `Edit in settings.json` like in screenshot above.

src/extension/package.json
jsondoo: Yea I agree, the settings was not really designed for this but from discussion with Stephane, we decided to put it here

src/extension/package.json
jsondoo: Also the docs are really limited and I haven't found an extension setting where they only show the strings yet

Tanya0609 (approved):

annielovesjs (approved):

jasmingoh (approved):

PR #1007, opened 2019-08-02
Added App Service Extension as dependency to our extension

jasmingoh:
This PR adds the Azure App Service Extension as a dependency to our extension.

We are doing this because we want to be able to do a simple `Azure App Service: Deploy to Web App` command in the generated project  to publish it to Azure.

**Test:**
Try to launch Web Template Studio without installing Azure App Service Extension, this error message will pop up:
image


SahilTara (approved):
Ayyy, also if you have a fresh install it automatically installs for you! nice

Tanya0609 (approved):

piotrmark (approved):

PR #1010, opened 2019-08-02
Accessibility task - titles, handling decorative svg

annielovesjs:
Changes
- hide decorative svg to screen reader
- changed text from octowizard to octopus wizard for screen reader, added screen reader text for add pages modal 

How to test
- download NVDA or [jaws](https://support.freedomscientific.com/Downloads/JAWS) screen reader

- when you go to an decorative svg, the screen reader shouldn't read anything (except for the dependency check icon - it should now read notification instead of icon for dependency checker for the screen reader


jcoc611-microsoft (approved):

Tanya0609 (approved):
Used JAWS, it works

streetcoder123 (approved):

PR #1014, opened 2019-08-06
App Service - create an instance

jasmingoh:
This PR handles the creation of an App Service web app.

To test:
- Launch the extension in preview mode
- Take note of the project name
- Add App Service on the Azure page (do "Create a new resource group")
- Create your project (try with different backend framework)
- Go to Azure portal (https://ms.portal.azure.com/#home)
    - Go to the Resource Group page and go into <project name>_<timestamp> resource group
    - Make sure that there is an ASP and a web app in the resource group
    - Go into the web app, then "Application Settings (Classic)", check that the Runetime stack corresponds to the backend framework 

Note: It is possible to create an instance in an existing resource group, just make sure that there is no Functions App in that resource group because we cannot have different kind of ASP in the same resource group.


src/extension/src/azure/azureServices.ts
```diff
@@ -354,19 +355,6 @@ export class AzureServices extends WizardServant {
     }
   }
 
-  // Will only be called if selections.resourceGroup is not an empty string
-  public static async createAppServicePlan(payload: any): Promise<string> {
```
jasmingoh, line 358: This is not required for now

src/extension/src/azure/azureServices.ts
```diff
@@ -437,6 +425,46 @@ export class AzureServices extends WizardServant {
     );
   }
 
+  public static async deployWebApp(payload: any): Promise<void> {
+    await AzureServices.updateAppServiceSubscriptionItemCache(
+      payload.appService.subscription
+    );
+    const aspName = await AzureServices.AzureAppServiceProvider.generateValidASPName(
```
jasmingoh, line 432: Not creating an ASP here because the SDK handles it

src/extension/src/azure/azure-app-service/appServiceProvider.ts
```diff
@@ -1,29 +1,80 @@
...
 
+  public async createWebApp(
+    selections: AppServiceSelections,
+    appPath: string
+  ): Promise<void> {
+    const template = this.getAppServiceARMTemplate();
+    const parameters = this.getAppServiceARMParameter(selections);
+    const deploymentParams = parameters.parameters;
+    try {
+      let options: ResourceManagementModels.Deployment = {
```
streetcoder123: Should we use `const` here?

src/extension/src/azure/azure-app-service/appServiceProvider.ts
```diff
@@ -92,6 +143,70 @@ export class AppServiceProvider {
...
+    let parameters = JSON.parse(fs.readFileSync(parameterPath, "utf8"));
+
+    parameters.parameters = {
+      name: {
+        value: selections.siteName
+      },
+      appServicePlanName: {
+        value: selections.appServicePlanName
+      },
+      sku: {
```
streetcoder123, line 177: Just out of curiosity, what is `sku`? or could you direct me to the documentation of the parameters?

src/extension/src/azure/azure-app-service/appServiceProvider.ts
```diff
@@ -118,8 +233,8 @@ export class AppServiceProvider {
     }
   }
 
-  private async generateValidASPName(name: string): Promise<string> {
-    let generatedName: string = NameGenerator.generateName(name, "asp");
+  public async generateValidASPName(name: string): Promise<string> {
```
streetcoder123, line 236: Why did you change the method from private to public? Is it being accessed out of this class?

src/extension/src/constants.ts
```diff
@@ -394,3 +395,8 @@ export enum OS {
   Linux = "linux",
   Windows = "windows"
 }
+
+export const BackendFrameworkLinuxVersion: { [s: string]: string } = {
+  Node: "node|lts",
+  Flask: "python|3.7"
```
streetcoder123, line 401: What are the versions used for? Do they mean that the app service would look for code that is compatible with this version of python?

src/extension/src/generationExperience.ts
```diff
@@ -146,7 +147,37 @@ export class GenerationExperience extends WizardServant {
 
     // Resource groups should be created before other deploy methods execute
     Promise.all(resourceGroupQueue).then(() => {
-      // Create ASP (this PR) and create web app (future PR) will be called here
+      if (payload.selectedAppService) {
+        serviceQueue.push(
+          GenerationExperience.Telemetry.callWithTelemetryAndCatchHandleErrors(
+            TelemetryEventName.AppServiceDeploy,
+            // tslint:disable-next-line: no-function-expression
+            async function(this: IActionContext): Promise<void> {
```
streetcoder123, line 155: Should we declare this function somewhere outside and then pass the function name into `callWithTelemetryAndCatchHandleErrors` for cleaner code?

src/extension/src/azure/azure-app-service/appServiceProvider.ts
jasmingoh: It is the pricing tier (https://docs.microsoft.com/en-us/azure/search/search-sku-tier)

src/extension/src/azure/azure-app-service/appServiceProvider.ts
jasmingoh: Yes, it is accessed out of this class now.

I initially thought that we need to create an App Service Plan for our user, but all that we need is a valid name. That's why we are using this method, and deleted this chunk: https://github.com/microsoft/WebTemplateStudio/pull/1014/files#diff-38b59b78a36d435c0373d85382d0d0e1L358-L368

src/extension/src/constants.ts
jasmingoh: version is for the runtime stack of the web app instance created

src/extension/src/generationExperience.ts
jasmingoh: will handle this in some other PR. I want things to be consistent with the way that we deploy functions and cosmos for now

streetcoder123: Okay!

streetcoder123 (approved):

jsondoo (approved):

annielovesjs (approved):

PR #1016, opened 2019-08-06
Improve screenreader experience for add pages and right sidebar

annielovesjs:
Changes/how to test:
- download[ nvda screen reader ](https://www.nvaccess.org/)
- when you go to the page count on add pages, it should read "# page(s)" instead of the number itself
- on the right sidebar, when you add multiple pages, the delete item icon should read "delete item" instead of cancel for the screen reader


jcoc611-microsoft (approved):

jsondoo (approved):
looks good

streetcoder123 (approved):

PR #1017, opened 2019-08-06
Autofill name field in Azure forms

### LINKED ISSUE #983: Autofill name field in Azure forms
Once a subscription is selected, it should autofill the name field with a valid name

jsondoo:
closes #983 

**How It Works**
- user opens Azure Services form
- when a subscription is selected, we send the project name to the extension
- extension returns a valid name to the client
- form is filled with the valid name

**How To Test**
- check that forms name fields are autofilled with a valid name on subscription change

image
(name is project name + unix timestamp)


src/client/src/reducers/azureLoginReducers/subscriptionDataReducer.ts
```diff
@@ -5,19 +5,22 @@ import AzureActionType from "../../actions/azureActions/azureActionType";
 {
     SubscriptionData: {
```
jasmingoh, line 8: should this just be removed?

src/extension/src/azure/azureServices.ts
```diff
@@ -210,10 +219,40 @@ export class AzureServices extends WizardServant {
...
+        );
+        break;
+      case AzureResourceType.Functions:
+        validationResult = await AzureServices.AzureFunctionProvider.checkFunctionAppName(
+          projectName,
+          userSubscriptionItem
+        );
+        break;
+    }
+    return validationResult ? false : true;
```
jasmingoh: this looks confusing.. just to clarify, the individual methods that checks if the name is valid return `undefined` if the name is valid, else, it would return an error message. so I was just trying to check if `validationResult` is `undefined`

src/client/src/reducers/azureLoginReducers/subscriptionDataReducer.ts
jsondoo: This just shows what the store would look like, and we have it in all the reducers so I think it's okay for now

src/extension/src/azure/azureServices.ts
jsondoo: maybe it's better to have `validationResult === undefined` then?

src/extension/src/azure/utils/nameGenerator.ts
```diff
@@ -1,12 +1,52 @@
...
     userProjectName: string,
-    resourceType?: string
+    resourceType?: string,
+    azureTypeName?: boolean
   ): string {
     const timestamp = unixToSuffix(Date.now());
     const suffix =
       resourceType === undefined ? timestamp : resourceType + "_" + timestamp;
-    return userProjectName + "_" + suffix;
+    return userProjectName + (azureTypeName ? "-" : "_") + suffix;
```
jasmingoh: wait, let's just change all `_` to `-`, it makes everything simpler

src/extension/src/azure/azureServices.ts
jasmingoh: that works too

jasmingoh:
not sure if this is a part of the buggy form, there was error was popping up very quickly

jasmingoh:
> not sure if this is a part of the buggy form, there was error was popping up very quickly

Talked with @jsondoo about it, and it is the existing form issue

jasmingoh (dismissed):

jsondoo:
> > not sure if this is a part of the buggy form, there was error was popping up very quickly
> 
> Talked with @jsondoo about it, and it is the existing form issue

Yea, similar behavior currently happens when you add a resource, and you click on Edit Resource, the validation error name will pop up for a split second - this can be addressed as a separate issue

annielovesjs (dismissed):

SahilTara:
> > not sure if this is a part of the buggy form, there was error was popping up very quickly
> 
> Talked with @jsondoo about it, and it is the existing form issue

One reason why the error shows up is allowing the name to have capital letters. If you look at the gif theres a split second where the name is `myApp-<some numbers>` instead of `myapp-<some numbers>`.

jasmingoh:
> not sure if this is a part of the buggy form, there was error was popping up very quickly
> 
> Talked with @jsondoo about it, and it is the existing form issue
> 
> One reason why the error shows up is allowing the name to have capital letters. If you look at the gif theres a split second where the name is myApp-<some numbers> instead of myapp-<some numbers>.

I noticed that too. @jsondoo added a this line (https://github.com/microsoft/WebTemplateStudio/pull/1017/files#diff-b6d9c6190ec2bb572aa7bd9c2854481aR17) to convert all the letters to lower case for Cosmos DB. So I would expect the Cosmos DB form to get a name with all lower case letters.. but it seems like that's not the case because of some bugs in the form

@jsondoo can you elaborate?

jsondoo:
@SahilTara @jasmingoh Okay, I think the bug in the gif you have is that we have one store (object) that stores all subscription related data. 

```
export interface ISubscriptionData {
  locations: any[];	
  resourceGroups: any[];	 
  validName: string;
}	
```

This `validName` field is used across all the forms (just like this subscription data is used across all forms), so when it initially opens the cosmos form, it loads in this existing valid name (if it exists from say opening the App Service form and selecting a subscription previously), and then since the time you select a different subscription in the cosmos form, it takes a few seconds to retrieve a new valid name that has lower case.

annielovesjs:
excellent work jason!

jsondoo:
@jasmingoh The bug in your gif should be fixed now. I decided to clear the `validName` field when the form is closed so that you don't see the valid names from other Azure forms. Turns out, doing this also helps make the code a bit cleaner :)

@Tanya0609 pointed out that the spinner is spinning in the name field even before a subscription is selected so this is hidden now. Also few more bug fixes

jasmingoh (approved):

streetcoder123 (approved):
Nice :)

Tanya0609:
I am on Mac. For me, when adding App Service, the validation alert did not show up, but when you click on Edit button, it shows. Jasmin also mentioned in her previous comment too. Is this bug addressable? If so, could you address this one too?

edit

Tanya0609 (commented):
Great Work 👍 But I have two questions. (I am on Mac.)

- [ ] For me, when adding App Service, the validation alert did not show up, but when you click on Edit button, it shows. Jasmin also mentioned in her previous comment too. Is this bug addressable? If so, could you address this one too?
edit

- [ ] When i add a new App service, right after I select a subscription group, the spinner starts spinning and then it populates with a SiteName. is this an ideal behaviour or an issue?
select

jsondoo:
> * [ ]  For me, when adding App Service, the validation alert did not show up, but when you click on Edit button, it shows. Jasmin also mentioned in her previous comment too. Is this bug addressable? If so, could you address this one too?

This one was existing behaviour/bug, I just looked into it and it's because when we save forms we are using `initialState` instead of using the `state` in our reducers (should be simple fix), so name availability was being set to the default `false`, so when the form is opened up again that shows up for a second. I can address it in a different PR 

Jasmin's bug was different, it was because I wasn't clearing `validName` so a valid name from a different form was showing up on other forms on open.

> * [ ]  When i add a new App service, right after I select a subscription group, the spinner starts spinning and then it populates with a SiteName. is this an ideal behaviour or an issue?

Yea that was behaviour I was going for, I set validation status to true as soon as the subscription is selected/changed. My concern is that there's a bit of a delay (few seconds) between selecting/changing subscription and the field being filled so I wanted some indication that it's loading something.

Tanya0609 (approved):

markAtMicrosoft (changes requested):
Unable to run this branch on my system.  We should figure out why it's not running before dumping into dev.

markAtMicrosoft (approved):

PR #1018, opened 2019-08-06
Add Warning for reset pages; update pages accordingly

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/32392

**Changes**
- Add a VSCode warning before resetting pages to confirm if the user wants to actually reset pages
- Do not redirect to pages page after resetting pages
- Update the page type count handler and a default blank page after resetting pages


src/extension/src/utils/vscodeUI.ts
```diff
@@ -33,7 +33,7 @@ export class VSCodeUI extends WizardServant {
   async promptUsersToResetPages(message: any) {
     if (
       message.payload.pagesLength > 0 &&
-      message.text === PAYLOAD_MESSAGES_TEXT.SWITCH_FRAMEWORKS_TEXT
+      message.text === PAYLOAD_MESSAGES_TEXT.RESET_PAGES_TEXT
```
jasmingoh, line 36: why? is this just outstanding from before?

streetcoder123: So, now we no longer need to reset pages on framework change. We reset the pages only if RESET_PAGES_TEXT is the payload for resetting pages. As of now, we could remove this text altogether, but it is good to have to distinguish between different commands if we want different behavior in future.

jasmingoh (dismissed):

src/client/src/App.tsx
```diff
@@ -249,26 +258,33 @@ class App extends React.Component<Props> {
...
+              {
+                title: "Blank",
+                internalName: `wts.Page.${selectedFrontend.internalName}.Blank`,
+                id: "Blank",
+                defaultName: "Blank",
+                isValidTitle: true,
+                licenses: [
+                  {
+                    text: "Bootstrap",
+                    url: "https://github.com/twbs/bootstrap/blob/master/LICENSE"
```
jcoc611-microsoft: Good candidate for `constants.js`?

src/client/src/App.tsx
```diff
@@ -249,26 +258,33 @@ class App extends React.Component<Props> {
...
+            // reset page count
+            const key = `wts.Page.${selectedFrontend.internalName}.Blank`;
+            const PAGE_TYPE_COUNT: IPageCount = {};
+            PAGE_TYPE_COUNT[key] = 1;
+            this.props.updatePageCount(PAGE_TYPE_COUNT);
+
+            // select default blank page
+            const PAGES_SELECTION: ISelected[] = [
+              {
+                title: "Blank",
```
jcoc611-microsoft, line 274: Does title need to be internationalized?

streetcoder123: Hmm, I am not sure. Does the pages that come from coreTS internationalized?

src/client/src/utils/constants.ts
```diff
@@ -30,6 +30,9 @@ const INTL_MESSAGES = defineMessages({
   }
 });
 
+const BOOTSTRAP_LICENSE =
+  "https://github.com/twbs/bootstrap/blob/master/LICENSE";
```
jsondoo, line 34: is there a reason this is not with the `const WEB_TEMPLATE_STUDIO_LINKS`?

jsondoo:
@njindal239 should we update this message? otherwise looks good
image

src/client/src/App.tsx
Tanya0609: The pages response from CoreTS are English only at the moment.

Tanya0609: Maybe instead of hard code the string, we could make it into a constant?

streetcoder123: I see, in that case it is not to be internationalized. It does not make sense to make the title a constant here since this is a default configuration that I have put inside of a variable :)

src/client/src/utils/constants.ts
streetcoder123: `WEB_TEMPLATE_STUDIO_LINKS` include the links that are related to the  WebTS repository. However, this is a license link for bootstrap. If more license links like these need to be included in future, we could group them under a licenses object.

jasmingoh (approved):

jsondoo (approved):

annielovesjs (approved):

PR #1019, opened 2019-08-06
Update regex for node version required

jasmingoh:
This PR is to address feedback we got about the dependency checker. Some people have Node version 12 installed, but they were still getting this warning (which is confusing):
image

Also, node version 12 will be active soon:
image

The regex is updated to match all node version above 10.15.


streetcoder123 (dismissed):
This won't break until Node version 20 comes out lol

jasmingoh:
> This won't break until Node version 20 comes out lol

it will be years from now lol

jcoc611-microsoft:
Since basically all dependencies would use semantic versioning, I think it would be better to use regex to extract the semantic version and then use [semver comparison](https://github.com/npm/node-semver) to test whether minimum version is met.

```javascript
const semver = require('semver');
const NODE_REGEX = RegExp('v(.+)')
const NODE_REQUIREMENT = '>=10.15.x';

// ...
let nodeVerison = 'v12.34.56'; // from node --version
let nodeSemver = nodeVersion.match(NODE_REGEX)[1];
let nodeMatches = semver.satisfies(nodeSemver, NODE_REQUIREMENT);
```

The current function is a bit more complex but has more edge cases that are not met.

SahilTara (approved):
Ayy added semver 😄 Nice!

streetcoder123 (approved):

jcoc611-microsoft (approved):
Nice!

PR #1022, opened 2019-08-06
hot fix for template tests

Tanya0609:
The reason the template tests failing was an unhandled error. This error supposed to be handled by an event in the coreTS script, but due to the event has the same name as some default name. The event doesn't work.  Thus this PR changed the event name to some new name and it solved the problem.

Also, fix two places based on tslint's suggestion in the test scripts.

**How to test:**
1. _cd src/extension_
2. run _yarn template-tests_, the tests should pass rather than throw an error

### DIFF
--- a/src/extension/src/coreTemplateStudio.ts
+++ b/src/extension/src/coreTemplateStudio.ts
@@ -105,9 +105,9 @@ export class CoreTemplateStudio {
     });
 
     process.stderr.on("data", (data) => {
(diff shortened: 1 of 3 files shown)


jcoc611-microsoft (approved):
:)

SahilTara (approved):
"hotfix" brings back memories

jasmingoh (approved):
ran the test, and it looks good
image

annielovesjs (approved):

PR #1024, opened 2019-08-06
Responsive home page on small screens

annielovesjs:
Changes:
Looked like this before:
notresponsive

looks like this now:
responsivehome


jsondoo (dismissed):
looks good

I think there's a bug with scrollbar and splash i found just now but filing it as a separate issue #1025

src/client/src/containers/NewProject/styles.module.css
```diff
@@ -46,3 +46,9 @@
 .quickStartContainer {
   margin-top: 1em;
 }
+
+@media (max-width: 1120px) {
+  .container {
+    max-width: 100%;
+  }
+}
```
streetcoder123, line 60: This would cause the input elements for `project name` and `save to location` to increase width to 100% on medium screens. Looks like this. Is this desired to be so long?
I would say maybe we could keep this to about 70% for medium screens and then to 100% for small screens.

image

annielovesjs: good point, will change it for medium screen sizes

streetcoder123 (approved):

Tanya0609:
On my Mac screen, regular size, I cannot see the title on the front page.


home

Tanya0609 (commented):
When in smaller screen, this is how it looks like in my local.
Screen Shot 2019-08-07 at 8 58 39 PM


Not sure if this is related to this PR. 
- On my Mac screen, regular size, I cannot see the title on the front page.

home

annielovesjs:
> When in smaller screen, this is how it looks like in my local.
> Screen Shot 2019-08-07 at 8 58 39 PM
> 
> Not sure if this is related to this PR.
> 
> * On my Mac screen, regular size, I cannot see the title on the front page.
> 
> home

This is actually fixed in my other PR. This PR's change was unrelated to the overlap

Tanya0609 (approved):

jsondoo (approved):
looks responsive

jcoc611-microsoft (approved):

Issue #1023, opened 2019-08-06
Improve post-generation modal to be scalable and more readable


xtinah-w:
![postGenerationModalScalable-v2](https://user-images.githubusercontent.com/24245311/62742924-55419c00-b9f5-11e9-8964-06a19752ddd7.png)

PR #1026, opened 2019-08-07
App Service - Postgen Modal

jsondoo:
**Changes**
- If app service is selected, it shows up on the post gen modal
- communication with extension happens through progressObject in #1014 

image

(our original plan was to not let users open project until app service is done, but we decided to just let them open it any time)

> If a user wants to use App Service, Functions, or Cosmos DB, they will be allowed to open the generated project at anytime, even if the resources haven't been deployed yet.
>  
> Reasoning: Users will want to edit their code first before doing any type of deployment. It is safe for them to poke through the code while this is happening.
>  
> Approach: Set expectations by letting users know how much time it will take to deploy each service (couple of minutes at worst)
>  
> 
> To follow this issue: https://github.com/microsoft/WebTemplateStudio/issues/1023


annielovesjs (approved):
looks great

Tanya0609 (approved):
works :) Good work!

src/client/src/containers/PostGenerationModal/index.tsx
```diff
@@ -106,7 +112,7 @@ const PostGenerationModal = ({
     if (isTemplatesFailed) {
       return formatMessage(messages.restartWizard);
```
jcoc611-microsoft: Should this be moved up before the `return`?

jcoc611-microsoft (approved):
Nice!

PR #1029, opened 2019-08-07
VueJS: updated through use of props.

ali-kha:
This PR is based on expert feedback. On this episode of Vue is Awesome, props! Props in vue are great because not only do they let you define their types with: 
```
props: {
  name: String
}
```
but they also let you define necessity:
```
props: {
  name: {
    type: String,
    required: true
  }
}
```
and that's not all! You can also add a custom validator:
```
props: {
  name: {
    type: String,
    required: true,
    validator: function(value) {
      if (value == "ali") return true;
      else return false;
    } 
  }
}
```
See you next time!

I went ahead and did this with all of our props to conform to the Vue Style Guide and the feedback from experts. To test: a newly generated Vue project should work as expected, and if a prop field is changed, for example if the project has a master detail page, in `Master_Detail.Vue` if you remove a field:
```diff
data() {
  return {
    masterDetailText: [
      {
        id: 0,
        longDescription: "",
        orderDate: "",
        orderTotal: 0,
-       shipTo: "",
        status: ""
      }
```
you should get a console error on the browser after running `yarn start`


templates/Web/Pages/Vue.List/src/components/ListItem.vue
```diff
@@ -23,7 +23,13 @@
 export default {
   name: "ListItem",
   props: {
-    listItem: Object
+    listItem: {
+      type: Object,
+      required: true,
+      validator: function(value) {
+        return !(value.text === undefined || value._id === undefined);
```
jcoc611-microsoft: Personally I think it's harder to read than `(value.text !== undefined || ...)`

ali-kha: @jsondoo

ali-kha: it would be `(value.text !== undefined &&` not `||` which is why I think this looks cleaner. Should we start a poll?

streetcoder123: I think the `value.text !== undefined && value._id !== undefined` is more readable but that's just a personal opinion.

templates/Web/Pages/Vue.MasterDetail/src/views/VueMasterDetail.vue
```diff
@@ -43,10 +43,13 @@ export default {
...
-          title: "",
-          tabName: "",
-          id: 0
+          id: 0,
+          longDescription: "",
+          orderDate: "",
+          orderTotal: 0,
+          shipTo: "",
+          status: "",
+          title: ""
```
streetcoder123, line 52: Why were all these parameters not already there? Was it even working without them?

templates/Web/Pages/Vue.MasterDetail/src/components/MasterDetailSideBarTab.vue
```diff
@@ -13,8 +13,14 @@
...
-    tabText: String,
-    index: Number
+    tabText: {
+      type: String,
+      required: true
+    },
+    index: {
+      type: Number,
+      required: true
+    }
```
streetcoder123, line 23: Out of curiosity - would the validation fail right away if we get some other type rather than `Number` for `index` prop?

templates/Web/Pages/Vue.MasterDetail/src/views/VueMasterDetail.vue
ali-kha: no, it was not working, this is pre-fetched data, and the fields were miss named so it was broken and I discovered it by writing the `required: true` in props.

templates/Web/Pages/Vue.MasterDetail/src/components/MasterDetailSideBarTab.vue
ali-kha: yes it would, it doesnt crash or stop rendering, but it throws a console.log error

templates/Web/Pages/Vue.List/src/components/ListItem.vue
jsondoo: yea `value.text !== undefined && value._id !== undefined` is more readable to me too

> deMorgans Law

jcoc611-microsoft (approved):

annielovesjs (approved):

streetcoder123 (approved):

PR #1033, opened 2019-08-07
Accessibility: Top nav bar overlap fixed

### LINKED ISSUE #991: On some screens the "Create Your Web App in Seconds" is cut off by Nav
On the Welcome page the Create Your Web App in Seconds gets cut off by the nav bar due to screen scaling.
 

image

Screen size + Scaling
image

annielovesjs:
Accessibility task:
- Increase text size to 200%. - nothing should be overlapping 

Changes/How to test:
- top nav bar should not overlap the title when you zoom to 200% on the first page

closes #991


streetcoder123 (commented):
The home page becomes like this when I zoom in

image

annielovesjs:
> The home page becomes like this when I zoom in
> 
> image

It should be good now! Forgot to add an additional change I made when I was testing on the browser xD

jasmingoh (dismissed):
Works!


I don't think people would zoom in this much and I'm not sure if this is an accessibility issue, but would it make more sense for the top nav bar to be on top of the footer?
image

jsondoo:
looks great, overlap is fixed, but i think quick starts needs margin/padding
image

annielovesjs:
> Works!
> 
> I don't think people would zoom in this much and I'm not sure if this is an accessibility issue, but would it make more sense for the top nav bar to be on top of the footer?
> image

We only have to consider up to 200% for accessibility :) so it is not an issue as of now

annielovesjs:
> looks great, overlap is fixed, but i think quick starts needs margin/padding
> image

Good catch! I added the fix

jsondoo (approved):
very nice :D

SahilTara (approved):

Tanya0609 (approved):

Tanya0609:
When it gets supper small, WebTS is overlaps with the step numbers. I think based on Annie's comment, this situation is not under consideration. Overall, great work!
Screen Shot 2019-08-08 at 10 57 13 PM

jcoc611-microsoft (approved):

PR #1044, opened 2019-08-08
VueJS: Cleaning up code to conform to the Vue Style Guide

ali-kha:
This PR is based on expert feedback. Based on the Vue Style Guide, I updated all of the attributes in the templates. for example, a data binding `:fooData` should go before an event listener `@barEvent`. This isn't a logic problem but a styling one. Also components that are used in multiple places for different purposes should begin with the word `Base`, so I added that to our Warning component. 

Because this is a purely styling update, to test, generating a Vue project should be exactly the same both in logic and UI.


jcoc611-microsoft (approved):

streetcoder123 (approved):

templates/Web/Pages/Vue.MasterDetail/src/components/MasterDetailSideBarTab.vue
```diff
@@ -1,10 +1,10 @@
 ﻿<template>
   <button
-    @click="$emit('onDisplayTabClick', index)"
-    type="button"
     class="list-group-item list-group-item-action styles.sidebarText"
+    type="button"
+    @click="$emit('onDisplayTabClick', index)"
   >
-    <img src="@/assets/GreyAvatar.svg" alt="Default Grey Avatar" class="mr-3" />
+    <img class="mr-3" src="@/assets/GreyAvatar.svg" alt="Default Grey Avatar" />
```
Tanya0609, line 7: Good work!

You move the class in Master Detail component, then you should move the class in Grid component too

ali-kha: I will be fixing it in the next VueJS PR

Tanya0609 (approved):

PR #1047, opened 2019-08-08
Azure Forms - Edit Resource

jsondoo:
**Changes**
Form behavior is changed so that after adding a resource, and then clicking 'Edit Resource', it should load the saved form data and enable the 'Save Changes' button (existing behavior is that it was disabled)
edit

(there's a few more remaining bugs in the form, some can't be repro'd consistently so I will file them after more investigation)


jasmingoh (approved):

annielovesjs (approved):
fantastic work :)

streetcoder123 (approved):

jcoc611-microsoft (approved):

PR #1050, opened 2019-08-08
Add Modals to Framework Details and Page Details Page

### LINKED ISSUE #1041: Licenses on learn more page don't show a modal confirming redirect like with view Licenses button.
Steps to repro:
1) Go to any "Learn more" page with licenses.
2) Click on it
3) Observe the instant redirect

image

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/32516

**Changes**
- Add redirect modals to details page for frameworks and pages as they redirect to non-microsoft websites.

closes #1041


src/client/src/containers/Details/index.tsx
```diff
@@ -28,24 +38,56 @@ interface IProps {
...
+  openRedirectModal
+}: Props) => {
+  const LinkRenderer = (props: any) => {
+    const genericRedirect: string = intl.formatMessage(
+      messages.genericRedirect
+    );
+    const redirectLinkLabel =
+      props.children[0].props.value === genericRedirect
+        ? props.href
+        : `${props.children[0].props.value} license link`;
```
streetcoder123, line 72: This was the best way I could think of to manage links that were done through 'React Markdown'. Since, the text is `here`, we cannot say the title where it is taking us. So the best bet was to mention the url of the site it is taking us to.

jasmingoh (dismissed):

jasmingoh (commented):
why did you move Details from components to containers?

streetcoder123:
> why did you move Details from components to containers?

Because it uses the Redux State now :)

annielovesjs (dismissed):

src/client/src/containers/Details/index.tsx
```diff
@@ -59,10 +101,13 @@ const Details = ({
...
   const renderFormattedData = (
     info: string | FormattedMessage.MessageDescriptor | undefined,
-    isMarkdown: boolean
+    isMarkdown: boolean,
+    isFrameworkDescription?: boolean
   ) => {
     if (formatteDetailInfo) {
       if (isMarkdown) {
+        if (isFrameworkDescription) {
+        }
```
jsondoo: Should this be here?

streetcoder123: Thanks for pointing that out. It should not be :)

src/client/src/containers/Details/index.tsx
jcoc611-microsoft: That's smart! I think `"${props.children[0].props.value} license link"` needs to be internationalized though, with `formatMessage`

src/client/src/containers/Details/index.tsx
```diff
@@ -154,13 +197,27 @@ const Details = ({
...
-                                  onKeyUp={keyUpHandler}
+                                <button
+                                  className={styles.licenseButton}
+                                  onClick={() =>
+                                    openRedirectModal({
+                                      redirectLink: String(licenseObject.url),
+                                      redirectLinkLabel: intl.formatMessage(
+                                        messages.redirectLabel,
+                                        {
+                                          redirectLabel: `${
```
jcoc611-microsoft, line 208: Same internationalization comment here

src/client/src/containers/Details/index.tsx
streetcoder123: This complete message constructed by redirectLinkLabel has been internationalized below before passing to the function.
`redirectLinkLabel: intl.formatMessage(messages.redirectLabel, {
              redirectLabel: redirectLinkLabel
            }),`

src/client/src/containers/Details/index.tsx
streetcoder123: This complete thing has been internationalized here. Everything is inside this:
`redirectLinkLabel: intl.formatMessage(`

jsondoo (dismissed):

src/client/src/containers/Details/index.tsx
```diff
@@ -1,6 +1,12 @@
 import classnames from "classnames";
 import * as React from "react";
 import ReactMarkdown from "react-markdown";
+import { connect } from "react-redux";
+import { ThunkDispatch } from "redux-thunk";
+
+import { AppState } from "../../reducers";
+import RootAction from "../../actions/ActionType";
+import * as ModalActions from "../../actions/modalActions/modalActions";
```
SahilTara: Small question. Do you really need to import * for this? 
Looks like you're only using openRedirectModalAction. Couldn't you do import { openRedirectModalAction } 
instead? Is there a reason why you're doing it like this rather than what I mention?

streetcoder123: Good point. Changed it.

jcoc611-microsoft (approved):

ali-kha (approved):

SahilTara (approved):

PR #1054, opened 2019-08-08
Accessibility: Communicate page state on screen reader

annielovesjs:
Before: 
- when you go to any of the links in the top nav bar, it just reads go to {page} page but does not communicate the state (ex. are we currently on {page}, or is {page} disabled?)

Changes:
- using NVDA screen reader, when the screen reader goes over the currently selected page, it should read current on {page}
- when the screen reader goes over a disabled page, it should read {page} disabled
- if the page is visited and not disabled or selected, it should just read {page} page


src/client/src/components/TopNavBarLink/index.tsx
```diff
@@ -39,20 +39,34 @@ const TopNavBarLink = ({
       onClick={handleClick}
       className={styles.container}
       onKeyUp={keyUpHandler}
+      aria-label={
+        visitedCheck || isSelected
```
jcoc611-microsoft, line 59: should we have `(..)`  parenthesis here?

SahilTara: Honestly I would extract some of this out into a const function or two. It is kinda hard to read/follow.

streetcoder123: @annielovesjs

annielovesjs: @jcoc611-microsoft the parenthesis gets automatically removed

annielovesjs: @SahilTara good point! I just abstracted it to a function call now so it should be more readable

SahilTara (approved):

jcoc611-microsoft (approved):

jsondoo (approved):
great work, tested with NVDA on windows

PR #1064, opened 2019-08-09
Remove button borders

jsondoo:
**Changes**
- Blue buttons are still blue w/o borders
- Other buttons are grey (same color as the quick start button) w/o borders

- Footer buttons are special case since it is a different background color - @xtinah-w and I discussed with the team and decided on the colors: back button should be black on dark mode, white on light mode (and checked this with accessibility compliance) 

**How To Test**
- check light theme and dark theme

image
image


streetcoder123 (dismissed):

jasmingoh (dismissed):

src/client/src/css/buttonStyles.module.css
```diff
@@ -1,14 +1,11 @@
 .buttonHighlightedBorder {
```
jcoc611-microsoft: Do we still need this rule? It is the same as `.buttonHighlighted` and it has no border (despite what name suggests)?

jsondoo: good catch, it is the same thing so I replaced it and removed it

annielovesjs:
doo-ing great

annielovesjs (approved):
doo-ing great

jcoc611-microsoft (approved):

ali-kha (approved):

PR #1065, opened 2019-08-09
Severity 2 Accessibility issue: add focus to top of page when navigating using the footer

annielovesjs:
Problem:
- on screen readers, when you go to the footer and click next or back, it takes you to a different page but the focus remains on the footer instead of top of the page
- when a visually impaired person tries to navigate using the down arrow (normal flow for screen readers), it takes you out of the wizard and reads the items from your other applications on the screen, preventing the user from completing the wizard 

Fix:
- when you navigate using the footer and hit back or next, the focus should go to the currently selected in the navigation bar

How to test:
- download NVDA screen reader
- go through the flow of the wizard and hit next or back on any of the pages and the item in focus should be the tab in the currently selected page on the navigation bar
- when you continue hitting the down arrow key, it should read the contents of the page in the wizard now


src/client/src/utils/constants.ts
```diff
@@ -40,6 +40,14 @@ const ARIA_LABELS_NAVIGATION = defineMessages({
   }
 });
 
+const PAGEID = {
```
SahilTara: Curious why don't we use string enum for this? Is there an advantage in terms of readability/type safety to do it this way rather than with enum strings?

jcoc611-microsoft (dismissed):

streetcoder123 (dismissed):
Great!

streetcoder123:
Now you have a conflict

src/client/src/utils/constants.ts
annielovesjs: Good point - it is better to use enum since these are ranged values. I just updated the PR to make it an enum instead of using a plain object

annielovesjs:
@njindal239  thanks for pinging me about the conflict! just resolved it :)

src/client/src/utils/constants.ts
SahilTara: Oh this update works too. I was thinking more like
```ts
enum SomeEnum {
  SomeVal="SomeString",
  SomeVal2="SomeString2"
}
```
Makes your code simple again while providing type safety.

jsondoo (approved):
good work 👍

piotrmark (approved):

jcoc611-microsoft (approved):

PR #1068, opened 2019-08-12
App Service - update application settings

jasmingoh:
This change is required to update the application setting of App Service web app.

When Cosmos DB is selected, we generate a .env file which contains the connection string to the database. If App Service and Cosmos DB are selected, we want the App Service application settings to also contains the connection string.

**To test:**
1. Launch the extension in preview mode
2. Add App Service and Cosmos DB (test for both Mongo and SQL)
3. Create the project
4. Once everything is deployed, click "Yes" in the prompt that updates the connection string in the .env file
5. Go onto Azure portal (https://ms.portal.azure.com/#home)
6. Go into the App Service created -> Configuration -> Application settings
image
7. Make sure it is the same as the content of the .env file


jsondoo:
Should my key in `.env` have two equals sign at the end for CosmosDB key?

jasmingoh:
@jsondoo There should be 2 equal signs for the key, I've updated the PR. Thanks for pointing it out! If you select a list page with App Service and Cosmos DB, you can successfully add stuff to the list.
Try here: https://myapp14-2019812164910.azurewebsites.net/List

Tanya0609 (approved):
Awesome, it works for SQL and MongDB!

piotrmark (approved):

jsondoo (approved):

PR #1069, opened 2019-08-12
Azure names - Zero-padded timestamp

jasmingoh:
This is required for generating consistent timestamps, i.e. yyyymmddhhmmss

The single digit component should be padded with 0.

**To test:**
1. Launch the extension in preview mode
2. Go to the Azure page
3. Add any Azure service to open up the forms
4. Select a subscription
5. Make sure tha the timestamp of the generated name follows the format above (yyyymmddhhmmss).

### DIFF
--- a/src/extension/src/azure/utils/nameGenerator.ts
+++ b/src/extension/src/azure/utils/nameGenerator.ts
@@ -61,9 +61,6 @@ export namespace NameGenerator {
   }
 
   function padStart(x: number): string {
-    if (x < 10) {
-      return "0" + x.toString();
-    }
-    return x.toString();
+    return x.toString().padStart(2, "0");
   }
 }
--- a/src/extension/src/azure/utils/nameGenerator.ts
+++ b/src/extension/src/azure/utils/nameGenerator.ts
@@ -62,7 +62,7 @@ export namespace NameGenerator {
 
   function padStart(x: number): string {
     if (x < 10) {
-      return 0 + x.toString();
+      return "0" + x.toString();
     }
     return x.toString();
   }
--- a/src/extension/src/azure/utils/nameGenerator.ts
+++ b/src/extension/src/azure/utils/nameGenerator.ts
@@ -52,11 +52,18 @@ export namespace NameGenerator {
     const fullDate = new Date(unixTimestamp);
     const year = fullDate.getFullYear().toString();
     // getMonth() returns month as a zero-based value
-    const month = (fullDate.getMonth() + 1).toString();
-    const date = fullDate.getDate().toString();
-    const hour = fullDate.getHours().toString();
-    const min = fullDate.getMinutes().toString();
-    const sec = fullDate.getSeconds().toString();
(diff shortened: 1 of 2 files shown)


src/extension/src/azure/utils/nameGenerator.ts
```diff
@@ -52,11 +52,18 @@ export namespace NameGenerator {
...
+    const date = padStart(fullDate.getDate());
+    const hour = padStart(fullDate.getHours());
+    const min = padStart(fullDate.getMinutes());
+    const sec = padStart(fullDate.getSeconds());
     return year.concat(month, date, hour, min, sec);
   }
+
+  function padStart(x: number): string {
+    if (x < 10) {
+      return 0 + x.toString();
```
jcoc611-microsoft: Should be `'0'` for readability. This still works as the number is implicitly converted to a string but when I read it at first it looked like math

src/extension/src/azure/utils/nameGenerator.ts
```diff
@@ -52,11 +52,18 @@ export namespace NameGenerator {
...
-    const sec = fullDate.getSeconds().toString();
+    const month = padStart(fullDate.getMonth() + 1);
+    const date = padStart(fullDate.getDate());
+    const hour = padStart(fullDate.getHours());
+    const min = padStart(fullDate.getMinutes());
+    const sec = padStart(fullDate.getSeconds());
     return year.concat(month, date, hour, min, sec);
   }
+
+  function padStart(x: number): string {
```
SahilTara, line 63: Doesn't a function for padStart already exist in ES? Although it's for strings so you'd have more .toStrings() haha. Your choice if you want to use it or not

jasmingoh: Yeap, I'm aware about that, but didn't want to do so many toString() just as you mentioned haha

SahilTara: Oh you could also get rid of the function block then and just replace it with a simple `return x.toString().padStart(2, "0")` if you want. But If not let me know it looks good otherwise.

jasmingoh: oh yea, that's a good idea

SahilTara (approved):

src/extension/tsconfig.json
```diff
@@ -3,7 +3,7 @@
     "module": "commonjs",
     "target": "es6",
     "outDir": "out",
-    "lib": ["es6", "dom"],
+    "lib": ["es2017", "dom"],
```
jsondoo, line 6: would this change possibly affect anything else?

jsondoo (approved):
nice :)

src/extension/tsconfig.json
jasmingoh: I changed this to use the padStart method, not sure if it will affect anything else.. @jcoc611-microsoft @piotrmark any idea?

annielovesjs (approved):

PR #1071, opened 2019-08-12
Azure names - trim generated name that are too long

jasmingoh:
This is required for always generating names (fields not exposed to the users) that are within the maximum length for each services.

**To test:**
1. Launch extension in preview mode
2. Enter a really long project name
3. Go to the Azure page and add any services
4. When the form opens up, error should show up and you'll have to shorten the name.
5. Add App Service to make sure that the generated name for App Service Plan (not exposed to user) is valid


src/extension/src/azure/utils/nameGenerator.ts
```diff
@@ -40,12 +40,22 @@ export namespace NameGenerator {
...
-    resourceType?: string
+    azureType?: AzureResourceType
   ): string {
+    let result;
     const timestamp = unixToSuffix(Date.now());
-    const suffix =
-      resourceType === undefined ? timestamp : resourceType + "-" + timestamp;
-    return userProjectName + "-" + suffix;
+    result = userProjectName + "-" + timestamp;
+    if (
```
SahilTara: Wondering if you should preserve the timestamp at all times since that is the uniqueness factor. Currently you strip away the timestamp if the name is too long! Think of the following situation: You select a CosmosDb name with length 31. You will be retrying the same name over and over again. That is not ideal. Consider another case where the length is around 27 you would be stripping away most of the date, hours minutes and seconds. 

Another idea would be to put a max length on the app name length as well!

jsondoo: I also think we should preserve the timestamp and trimming the project name

jasmingoh: @SahilTara thanks for pointing it out, it totally make sense

src/extension/src/azure/azure-app-service/appServiceProvider.ts
```diff
@@ -65,9 +65,12 @@ export class AppServiceProvider {
         selections.subscriptionItem
       );
       this.writeARMTemplatesToApp(appPath, template, parameters);
+      const deploymentName = (
+        selections.siteName + APP_SERVICE_DEPLOYMENT_SUFFIX
+      ).substr(0, CONSTANTS.DEPLOYMENT_NAME.MAX_LENGTH);
```
jasmingoh, line 71: constraint of deployment name is 64 characters that's why I'm trimming it (also, since deployment names are not very visible for users, I just did a simple trim -- I could trim the project name, leave the timestamp in tact, but I just feel like that's too much logic for generating a deployment name)

Tanya0609 (dismissed):
It works! I have "myApp10asdfffsdgdfsgdsfbgdsfgkvjafdgjnsdgdnsfdsfn" as project name, and this is the generated name.
Screen Shot 2019-08-12 at 9 34 19 PM

src/extension/src/constants.ts
```diff
@@ -260,6 +260,16 @@ export const CONSTANTS = {
     MAX_LENGTH: 60,
     MIN_LENGTH: 3
   },
+  ASP_NAME: {
```
jcoc611-microsoft: Easily confused with ASP.NET :P

jasmingoh: do you suggest to change the constant name ?

jasmingoh:
@Tanya0609 can you pull and try it again? the error didn't show up when I did it..

src/extension/src/azure/utils/nameGenerator.ts
```diff
@@ -16,36 +15,30 @@ export namespace NameGenerator {
...
+    const suffix: string = "-" + timestamp;
+    return trimProjectName(userProjectName, azureType) + suffix;
+  }
+
+  function trimProjectName(
+    projectName: string,
+    azureType?: AzureResourceType
+  ): string {
+    const suffixLength = 15; // format of suffix is "yyyymmddhhmmss-"
+    if (azureType === AzureResourceType.AppServicePlan) {
```
jasmingoh, line 35: this is still required since we do not expose this field to our user.

jasmingoh:
Talked with @Tanya0609 and @xtinah-w and we felt that the appropriate way of handling this is to not trim the project name for our users and show the error message instead

xtinah-w:
Yeah, please don't trim names because that's unexpected behaviour for the user to have to deal with and if they put in a 50 character name, chances are, they put a lot of thought into it and it would not be a great experience to have to erase stuff and then type it back in. Most users would want to remove the timestamp!

streetcoder123 (dismissed):

src/extension/src/azure/utils/nameGenerator.ts
```diff
@@ -16,36 +15,30 @@ export namespace NameGenerator {
...
-    return userProjectName + "-" + suffix;
+    const suffix: string = "-" + timestamp;
+    return trimProjectName(userProjectName, azureType) + suffix;
+  }
+
+  function trimProjectName(
+    projectName: string,
+    azureType?: AzureResourceType
+  ): string {
+    const suffixLength = 15; // format of suffix is "yyyymmddhhmmss-"
```
ali-kha: NIT: the format comment is wrong, it should be `... is "-yyyymmddhhmmss"`

annielovesjs (dismissed):

src/extension/src/azure/utils/nameGenerator.ts
jasmingoh: thank you

ali-kha (approved):

annielovesjs (approved):

streetcoder123 (approved):

PR #1075, opened 2019-08-12
Add build scripts for all front-end frameworks

streetcoder123:
Task: https://microsoftgarage.visualstudio.com/Intern%20GitHub/_workitems/edit/32678

**Changes**
- Add a script for the command `npm run build`
- The script is added to make the `mv` command cross-platform
- Also, it checks for corner cases such as overwriting the build folder in case it is already present in the server folder.


jasmingoh (approved):
tested on mac - works for react, angular and vue!

piotrmark (approved):

ali-kha (approved):

PR #1087, opened 2019-08-14
hot fix on vue flask grid on mac

Tanya0609:
CoreTS supposes to combining files based on alphabetical order. When combining constant.py, it supports to insert an empty array, then inserts the value to this array.  However, coreTS did not follow the order it supposed to do. Thus, renaming one of the file reinforces the order and solves the issue.
 
Thanks @SahilTara for debugging to find the cause and coming up with this change.


templates/Web/_composition/Flask/Page.TheFlask.Grid.AddMerging/.template.config/template.json
```diff
@@ -0,0 +1,28 @@
+{
+  "author": "Microsoft Community",
+  "classifications": [
+    "Universal"
+  ],
+  "name": "Page.Flask.Grid.AddMerging",
```
streetcoder123: You need to change this to `Page.TheFlask.Grid.AddMerging`

streetcoder123 (changes requested):

templates/Web/_composition/Flask/Page.TheFlask.Grid.AddMerging/.template.config/template.json
Tanya0609: I was thinking that one too. Thanks for the catch! PR updated.

SahilTara (approved):

jasmingoh (commented):
Tested all Flask applications (with all pages) on Mac and Linux and they worked

jsondoo (approved):

streetcoder123 (approved):

annielovesjs (approved):

Issue #1111, opened 2019-08-19
Improve Azure Services experience based on user research feedback


xtinah-w:
![newAzureServicesFlow](https://user-images.githubusercontent.com/24245311/63288996-e70d9c80-c272-11e9-999a-56330676601b.png)
![azureServicesFlow-master](https://user-images.githubusercontent.com/24245311/63288560-df012d00-c271-11e9-9fc3-1d49082d5b37.png)
After conducting user testing with intermediate web app developers, the current existing design still doesn't fulfill the knowledge needed to feel confident in adding Azure services to our product. 

Our team has determined that the best way to handle this is through text and visual feedback via a video embedded into the wizard. We also believe a seamless wizard experience is necessary to reduce the friction from adding a service and to create a clear picture of the relationship between the Azure log in to add the Azure services by setting expectations (as shown in the log in/sign up modal) for the user.

### **Task List**

**Azure Sign In Page**
- [x] Create log in/create account modal - @Tanya0609 (#1121)
- [x] Re-route "Add Resource" button to payment modal - @Tanya0609 (#1137)

**Azure Signed In Page**
- [x] Make more apparent you are signed in (#1121)

**Azure Sign In/Signed In Page**
- [x] Add payment & time expectation - @jasmingoh (#1113)
- [x] ~Investigate how to add videos to the wizard~ - @jasmingoh
- [x] ~Create App Service video~
- [x] ~Create Cosmos DB video~
- [ ] Ensure responsiveness of page (reference: https://cloud.google.com/)
- [x] Add section where user is logged in (#1121)
- [x] Remove Azure Functions from preview mode - @jasmingoh (#1122)

**Log In/Create Account Modal**
- [x] Remove Azure Sign In Section - @Tanya0609  (#1137)
- [x] Create copy for log in/create account modal
- [x] Create dropdown behaviour for FAQ to set expectations for user (#1121)

**Azure Forms**
- [ ] If user selects, "Create New" for the subscription part, the entire wizard should be disabled apart from a single refresh function (with option to cancel) so that the user doesn't have to exit out of the form then add it back in
- [x] default "create a new resource group" option and hide resource group section in both forms, since the regular user flow is to put all services related to one project into the same resource group. Here we are creating a new project for them, and create a new resource group too, thus no need to provide "choose existing" option - @jasmingoh (#1117)
[...]

PR #1113, opened 2019-08-19
Azure - Add time and money expectation

jasmingoh:
This is required for #1111 (Add payment & time expectation)

**@xtinah-w's mock:**
image

**Before:**
image
image

**After:**
image
image


jcoc611-microsoft:
I'm guessing there is no Azure API for getting the current values of these offerings? If not, do we need to run this through CELA for approval?

src/client/src/mockData/azureServiceOptions.ts
```diff
@@ -160,6 +159,17 @@ export const servicesEnum = {
...
 };
 
+export const servicesExpectation = {
+  APP_SERVICE: {
+    price: "Free 30 Day Trial",
+    time: "Under 3 minute set-up time"
+  },
+  COSMOS_DB: {
+    price: "12 months of free access",
+    time: "Under 10 minutes set-up time"
```
jsondoo: should we localize these strings

ali-kha:
Probably should add aria labels and alt texts to the SVGs

jasmingoh:
@jcoc611-microsoft we will bring it up during our sponsor sync meeting tomorrow, thanks for pointing it out

src/client/src/types/option.d.ts
```diff
@@ -17,4 +17,6 @@ export interface IOption {
   unselectable?: boolean;
   isValidTitle?: boolean;
   author?: string;
+  expectedTime?: string | FM.MessageDescriptor;
+  expectedPrice?: string | FM.MessageDescriptor;
```
streetcoder123, line 21: What is FM.MessageDescriptor?

jasmingoh: oh, just look at line 4!

src/client/src/mockData/azureServiceOptions.ts
```diff
@@ -59,6 +59,14 @@ export const azureMessages = defineMessages({
     defaultMessage:
       "Connect your web app to a distributed database service to access and query data using SQL or MongoDB API."
   },
+  cosmosTimeExpectation: {
+    id: "cosmosDb.timeExpectation",
+    defaultMessage: "12 months of free access"
```
SahilTara, line 64: It's not 12 months of free access forever though right (see later sentence for explanation of what I mean)? And the free access has limits right? Ignoring everything I know about Azure, I think this message conveys that I can use as much of cosmos  possibly even 10 Gb-1 Tb of storage for free for 12 months, which I don't think is true. Also may convey that every time the user creates a new instance they get 12 month's free which I'm not sure is true either. Correct me if i'm wrong here. I think you should have some more clear wording for these things or an annotation saying that it's only free from the first instance's launch up to a certain point, also linking the hard limits per month if any.

jasmingoh:
@jcoc611-microsoft we've just talked to our sponsors about it. App Service has been offering the 30 day free trial for as long as they remember. As for Cosmos DB, I'll work with @xtinah-w to figure out better wordings such that it is not misleading. @SahilTara

Tanya0609:
cosmos-order

jasmingoh:
please don't review this first - still waiting for updated design

PR #1115, opened 2019-08-20
Project Path Validation should Fail on emojis

### LINKED ISSUE #1103: Having emojis in 'Save To' path, then generating and opening project fails
Steps To Reproduce
1) Create a new folder with unicode characters in the folder name
2) Select that folder to be your 'Save To' path
3) Click 'Create Project' then 'Open Project'

image

streetcoder123:
GitHub Issue #1103 

**Changes**
- If we create a folder with emojis and assign it to 'Save To' location, it was not giving any validation error and allowed the user to proceed. However, generation of project in that case failed.
- We need to disable a project path that contains emojis.

closes #1103


SahilTara (approved):

src/extension/src/utils/validator.ts
```diff
@@ -74,6 +74,14 @@ export class Validator extends WizardServant {
     let isValid = true;
     let error = "";
 
+    if (
+      /(?:[\u2700-\u27bf]|(?:\ud83c[\udde6-\uddff]){2}|[\ud800-\udbff][\udc00-\udfff]|[\u0023-\u0039]\ufe0f?\u20e3|\u3299|\u3297|\u303d|\u3030|\u24c2|\ud83c[\udd70-\udd71]|\ud83c[\udd7e-\udd7f]|\ud83c\udd8e|\ud83c[\udd91-\udd9a]|\ud83c[\udde6-\uddff]|[\ud83c[\ude01-\ude02]|\ud83c\ude1a|\ud83c\ude2f|[\ud83c[\ude32-\ude3a]|[\ud83c[\ude50-\ude51]|\u203c|\u2049|[\u25aa-\u25ab]|\u25b6|\u25c0|[\u25fb-\u25fe]|\u00a9|\u00ae|\u2122|\u2139|\ud83c\udc04|[\u2600-\u26FF]|\u2b05|\u2b06|\u2b07|\u2b1b|\u2b1c|\u2b50|\u2b55|\u231a|\u231b|\u2328|\u23cf|[\u23e9-\u23f3]|[\u23f8-\u23fa]|\ud83c\udccf|\u2934|\u2935|[\u2190-\u21ff])/g.test(
+        path
```
jsondoo, line 79: Should we just check if the path is alphanumeric or something (like we do with the project name)? 'Open Project' will still fail for other unicode characters

streetcoder123: I agree but I am not sure if we should/can allow other symbols like quotation marks, spaces in path etc. What do you think?

jsondoo (approved):

piotrmark (approved):

PR #1116, opened 2019-08-20
Removed orange highlight from selections onClick but kept for tab

### LINKED ISSUE #1086: Remove yellow highlight on selection
Causes too much confusion and uncertainty which we've gotten feedback on during user testing

SahilTara:
To Test:
- Confirm when not currently tabbed the selections (Frameworks + Pages) keep a white outline.
- Confirm when tabbed they become orange.
- Confirm that all cards (SelectableCards) haven't lost any intended highlight, whether it be white or 
orange!

Changes:
- Added polyfill for focus-visible, and used its feature to have selective orange highlight.

@xtinah-w is this what you wanted with #1086 ?  (Build my branch, or just run the client after going into my branch).
closes #1086


Tanya0609:
Instead of using a package, will there be a solution that overwrites the vs code default colour? in the src/extension/package.json file 

More docs:
https://code.visualstudio.com/api/references/contribution-points#contributes.colors
https://code.visualstudio.com/api/references/theme-color#editor-colors

SahilTara:
> Instead of using a package, will there be a solution that overwrites the vs code default colour? in the src/extension/package.json file
> 
> More docs:
> https://code.visualstudio.com/api/references/contribution-points#contributes.colors
> https://code.visualstudio.com/api/references/theme-color#editor-colors

Hey @Tanya0609 the reason I'm using the package is for easy overriding of the behaviour of clicking vs tabbing. Although in #1086 it is also mentioned that we should change the colour to blue. Is that what you're referring to in the comment? I'm open to any suggestions for possible colours. My main concern is the contrast with our call to actions. Any ideas?

xtinah-w:
I'll check this in the morning, thanks Sahil!! This was really bothering me since this feedback has existed since the beginning of our internship.

SahilTara:
@xtinah-w any updates?

jsondoo:
Is there any compliance things we need to be aware of if we're adding a new package?

jsondoo (approved):

jsondoo (approved):
@xtinah-w also approves

SahilTara:
> Is there any compliance things we need to be aware of if we're adding a new package?

Just have to add it to the open source package registration if anything. For subscription into bug notifications and such. Just name and version number I'm pretty sure.

xtinah-w (approved):
Thanks, Sahil!

piotrmark (approved):

PR #1117, opened 2019-08-20
Azure - simplify forms

jasmingoh:
After talking with our users and sponsors, we decided to remove the resource group field from both Cosmos DB and App Service forms, and remove the location field from the Cosmos DB form.

**Before**:
image
image
___
**After**:
image
image

___
**To test:**
- Create a project, add Cosmos DB and App Service
- After deployment is done, go to the portal and find your resource group
- Make sure that the Cosmos DB and App Service is in the resource group
- Both services should be in Central US


jsondoo (commented):
We should also probably remove `chooseExistingRadioButtonSelected` from the initial states, React hooks and handlers (i.e. `radioButtonOnChangeHandler`) for the radio buttons.

src/client/src/containers/AppServiceModal/index.tsx
```diff
@@ -259,7 +256,7 @@ const AppServiceModal = (props: Props) => {
       siteNameAvailability,
       setFormIsSendable
     );
-  }, [appServiceFormData.chooseExistingRadioButtonSelected]);
+  });
```
jsondoo: we should remove this `useEffect` completely now

Tanya0609 (approved):
The form is much simpler now!

piotrmark (approved):

jcoc611-microsoft (approved):

SahilTara (approved):
I now understand what was being removed and why. Nice!

PR #1122, opened 2019-08-21
Azure - Remove Functions completely from UI

jasmingoh:
With the new Azure page design, we decided to take Functions out of preview mode as well.

All the extension side code will remain in the code base, except for the top layer (i.e. the generationExperience)


SahilTara:
So no more Azure functions? Might wanna adjust the wording since it seems like you are setting preview to false but that's not the case right? You seem to be removing functions entirely from the UI.

Tanya0609 (approved):
Functions is no longer show in the frontend!

Tanya0609 (changes requested):
When I tried to generate a project with App Service and CosmosDB, it crushes at post generation modal

jasmingoh:
@Tanya0609 it's fixed! please try to break it and see if there is other bugs

SahilTara (approved):

piotrmark (approved):

jcoc611-microsoft (approved):

jsondoo (approved):

Tanya0609 (approved):

PR #1130, opened 2019-08-22
Possible fix for right sidebar closing

### LINKED ISSUE #1105: Right Sidebar closes abruptly when deleting pages
Steps to Reproduce:
- Add a few pages to an application
- Remove pages from the right sidebar
- As soon as 1 page is left, the sidebar closes abruptly
- Then, as you add a page, it abruptly appears again.

SahilTara:
Hey All this PR attempts to fix #1105.

Any thoughts?
We now persist the sidebar as long as a page change or azure addition has happened.

### DIFF
--- a/src/client/src/containers/RightSidebar/index.tsx
+++ b/src/client/src/containers/RightSidebar/index.tsx
@@ -135,8 +135,9 @@ class RightSidebar extends React.Component<Props, IRightSidebarState> {
     if (!prevState.isSidebarUserControlled) {
       return {
         isSidebarOpen:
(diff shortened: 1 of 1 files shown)


piotrmark (approved):
Nice! I was thinking about setting isSidebarUserControlled to true when it automatically opens the first time (user added page or service), but since this is another component, your solution is definitely less complicated :)

jsondoo (approved):

streetcoder123 (approved):
Looks Good :)

PR #1151, opened 2019-08-25
Flask Template - Port to 5000 for Flask Standard

Arcaster42:
Flask works out-of-the-box on port 5000 but not on other ports (such as the 3001 used), so starting a project will lead to issues and components not working (as I experienced with Vue). This is alleviated by leaving the Flask port to 5000 and allowing the user to configure further as desired. Ultimately, we don't want end-users to have a bundled project that doesn't work on the first start, so I'm making this PR.

### DIFF
--- a/templates/Web/Projects/FlaskDefault/server/constants.py
+++ b/templates/Web/Projects/FlaskDefault/server/constants.py
@@ -1,7 +1,7 @@
 import os
 
 CONSTANTS = {
-    'PORT': os.environ.get('PORT', 3001),
+    'PORT': os.environ.get('PORT', 5000),
     'HTTP_STATUS': {
         '404_NOT_FOUND': 404,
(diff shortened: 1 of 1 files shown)


SahilTara:
Hey thanks for the contribution,

I have a few questions:

Which os does this not work on?

Have you tried pip installing flask manually, if `yarn install-dependencies` fails and then running?

I am trying to get this info because I cannot seem to reproduce the issue at the moment.

Arcaster42:
This was on Windows 10. Doing a manual reinstall corrected the issue for me. It seems like it must have been on my end. Thanks for your time.

SahilTara:
Thanks for bringing up the issue though. Having to manually install doesn't seem to be intuitive. We will attempt to make this a bit easier in a future release!

PR #1152, opened 2019-08-25
Fixed python dependency checker for Anaconda3 5.2.0

### LINKED ISSUE #1150: Python installation not detected
Python is not detected, and I get an error: `Error Generating templates`. The folder created by Web Template studio doesn't work when I run it because of this.

Screenshot from 2019-08-25 15-57-28

Screenshot from 2019-08-25 15-59-26

SahilTara:
Summary of changes:
- Added handling for stripping characters after the python version.
- Utilized stderr output since older versions of anaconda used stderr for output instead of stdout.

closes #1150


src/extension/src/utils/dependencyChecker.ts
```diff
@@ -31,15 +31,27 @@ export class DependencyChecker extends WizardServant {
   private async runPythonVersionCommand(command: string) {
     let installed: boolean;
     try {
-      const { stdout } = await exec(command + " --version");
-      const version = stdout.match(PYTHON_REGEX)[1];
+      const { stdout, stderr} = await exec(command + " --version");
+      // stderr is also processed for older versions of anaconda!
+      const matches = stdout.match(PYTHON_REGEX) || stderr.match(PYTHON_REGEX);
+      const version: string = this.getVersionFromMatch(matches[1]);
```
jcoc611-microsoft: I think you could just modify the PYTHON_REGEX instead of using `getVersionFromMatch`, maybe something like `RegExp("Python ([0-9.]+)")`

SahilTara: Hmm, yeah I was thinking that as well since other possible matches would be in different indicies. I'll try it out!

SahilTara: Worked! Updated the code accordingly.

jcoc611-microsoft (approved):

Tanya0609:
@SahilTara How can I test the code? Will I need to download anaconda?

SahilTara:
> @SahilTara How can I test the code? Will I need to download anaconda?

Yup download anaconda3 5.2.0. Ensure that when you type python or python3 it launches anaconda's python 3.6.5. Ensure that on the dev branch python isnt detected, but on my branch it will be detected. Its easier to test on mac/linux but you can still test it on windows I think, might have to update PATH manually though.

Tanya0609:
Also, just curious, for people like me, who has python 2.x downloaded, then have Anaconda3 5.2.0 downloaded. What will happen?

After I downloaded Anaconda3, when i type in python, it still shows 2.x and when i type python3, it shows not valid. Did I do something wrong?

SahilTara:
> Also, just curious, for people like me, who has python 2.x downloaded, then have Anaconda3 5.2.0 downloaded. What will happen?
> 
> After I downloaded Anaconda3, when i type in python, it still shows 2.x and when i type python3, it shows not valid. Did I do something wrong?

Are you on mac? It probably asked you to allow it to modify your bashrc and you might have hit no because that is the default.
Add this line to your ~/.bashrc
```
export PATH="~/anaconda3/bin:$PATH"
```
Or wherever anaconda3/bin is located.
Not sure if you need the exact path or if the ~ is alright here.

Then save the file and do `source ~/.bashrc`
Then try typing python again. And python3. Both should work. If not try using the absolute path or restarting the terminal session.

Python 2.x is hidden by anaconda's python 3.6 5. They use the same alias but anaconda's will come first.

jsondoo (approved):
works :)

Tanya0609 (approved):

Issue #1123, opened 2019-08-21
Angular Feedback (New)


Tanya0609:
## Angular Feedback from John Papa 
Although the comments are documented based on each file, it apply to the entire generated project

### one liner / remove unnecessary comments and lines
- [x] Warning-message.module.ts: Make the arrays all one liners (smaller is better) 
- [x] Removing extra lines makes generated code smaller (usually people like this) 
- [ ] Master-detail.service.ts: @Injectable({}) should be one line 
- [x] App.module.ts: Empty arrays, objects should be removed  
- [x] App.component.html: Remove comments, no extra stuff is necessary 
                   - Every comment should be intentional and tell something really important 
                   - If the comment wasn't there, would it still make sense? 
- [x] app-routing.module.ts: remove empty objects from NgModule 
                  - When you generate pages, it might be nice to have a home page 

### code styles
- [x] Newer ones do not use do providers array, should be moved to service 
- [x] change mod to m or module
- [x] function or variable names should be camel case i.e. master-detail.component.ts
- [x] Interface should have its own file 

### add error handling
- [x] Add in error handling (e.g. this.http.get should have a catch) 

### environments
- [x] environments: add a comment to both prod and dev file to explain the difference between the two

Tanya0609:
## Server related code

- [  ] BuildScript.js 
- Put a comment at the top explaining what this script is for 
- People may be confused why angular-cli is not being used 

-[ ] Package.json
- Don't use nodemon, you don't want that to go to production 
- Take out nodemon completely and just use node 
- Right now nodemon is a devDependencies, this may break things when it's deployed

Tanya0609:
### Accessibility 
- [x] Accessibility standards for Angular code should be considered (for example, alt text on images, aria role) 

### CSS
- [x] If necessary, clean up app.component.css. The CSS here will be inherited by all the children.
- [ ] double check the way we import svgs, is it the way to import svgs in css? If not, then fix it.

Tanya0609:
- [  ] Environments 
- Add a comment to both prod and dev file to explain the difference between the two 

- [  ] Main.ts 
- Looks good 😊 
- Consider looking into progressive apps (PWA)

Tanya0609:
Although the comments are documented based on each file, it applys to the entire generated project

mcmxcdev:
@Tanya0609, I would like to work on this, since I have a couple years experience with Angular and worked with Angular 2-8 and know a thing or two about best practices ;)

[...]

PR #1157, opened 2019-08-27
Angular Feedback (New)

mcmxcdev:
Started working on improving the generated Angular project and this is intended as a WIP draft to see if the direction is right.

Since the referenced issue #1123 is very big, I would like to get some feedback on which checkboxes from the issue I should work on in this PR and which ones will be delegated to other issues/PRs.


.gitignore
```diff
@@ -338,6 +341,7 @@ src/extension/package.nls.*.json
 **/build/
 **/package-lock.json
 **/*.vsix
+src/extension/.vscode-test
```
Tanya0609: Just curious, is there a reason you add this file into .gitignore?

Tanya0609:
Hi @mhatvan, great work! You are definitely in the right direction. I have split #1123 into smaller issues for you, please take a look and let me know if you have more questions. 

I have copied two groups of feedback from #1123, which you have achieved majority in this PR already. In my opinion, it will be great if you could help with those feedback in this PR. However, feel free to determine the scope of the PR too.

### one liner / remove unnecessary comments and lines
- [ ] Warning-message.module.ts: Make the arrays all one liners (smaller is better) 
- [ ] Removing extra lines makes generated code smaller (usually people like this) 
- [ ] Master-detail.service.ts: @Injectable({}) should be one line 
- [ ] App.module.ts: Empty arrays, objects should be removed  
- [ ] App.component.html: Remove comments, no extra stuff is necessary 
                   - Every comment should be intentional and tell something really important 
                   - If the comment wasn't there, would it still make sense? 
- [ ] app-routing.module.ts: remove empty objects from NgModule 
                  - When you generate pages, it might be nice to have a home page 

### code styles
- [x] Newer ones do not use do providers array, should be moved to service 
- [x] change mod to m or module
- [x] function or variable names should be camel case i.e. master-detail.component.ts
- [x] Interface should have its own file

.gitignore
mcmxcdev: I added `src/extension/.vscode-test` & `src/extension/src/template_test` to .gitignore because they both dynamically generate files which don't need to be version controlled, similar to a `/dist` folder

mcmxcdev:
I resolved a merge conflict in `.gitignore`, @sibille can you take a look at this and get it merged?

sibille:
Thanks @mhatvan, we'll review and merge this week! Sorry again for the delay!

sibille:
Great work @mhatvan, thanks for your contribution!

sibille (approved):

PR #1173, opened 2019-09-06
Adding MoleculerJS #1124

AndreMaz:
This PR is an attempt to add [Moleculer](https://moleculer.services/) to the list of back-end frameworks https://github.com/microsoft/WebTemplateStudio/issues/1124

What was done/changed:
- Default Template containing:
  - 2 Moleculer microservices
    - HTTP gateway service. *Note*: 404 handler for nonexistent endpoints is present by default
    - `Pages` service containing the application logic
  - `Readme.md` + links pointing to Moleculer docs for more info
  - Fully documented `moleculer.config.js` for backend tweaking
- Composition Templates for the following requests:
  - `GET /grid`
  - `GET /list`
  - `POST /list`
  - `DELETE /list/:_id`
  - `GET /masterdetail`
- Client updates
  - Had to update some files in `src/client/src` to add the `.svg` icon to the Wizard

What's not done:
- Integration with Azure Cosmos

Some prints (copy&paste from #1124)
Wizard: 
image
Moleculer+Vue
FM2eNnx8T4

PR #1183, opened 2019-09-19
[moleculer] template improvements

AndreMaz:
Template https://github.com/microsoft/WebTemplateStudio/pull/1173 improvements + CosmosDB composition templates + Refac based on @icebob's feedback :
- Composition templates for Mongo and SQL APIs
- Debug config
- Added params validation
- More docs
- started() and stopped() methods are `async`
- Refactored `DELETE /list/:_id`. Initial version was copy+paste from Express template
- preview: false (see https://github.com/microsoft/WebTemplateStudio/pull/1183#issuecomment-540503547)
- Refactored all templates as per @icebob's (creator of Moleculer) feedback (see https://github.com/microsoft/WebTemplateStudio/pull/1183#issuecomment-540151694 and https://github.com/microsoft/WebTemplateStudio/pull/1183#issuecomment-540496691)


sibille:
@AndreMaz is this ready for review or still in progress?

AndreMaz:
@sibille it's ready for review. ~This PR just contains a few tweaks and a debug config.~

AndreMaz:
@sibille I've updated this PR. It now contains CosmosDB composition template. 

PS: It's ready for review

AndreMaz:
Ok. So now this PR also contains composition templates for Mongo and SQL APIs.

@sibille @dgomezc after https://github.com/microsoft/WebTemplateStudio/pull/1173 and this PR, what's necessary for Moleculer to be out of preview mode (https://github.com/microsoft/WebTemplateStudio/commit/c09d6307b959b4ae9573d06b7bec4f2df9514a40)?

dgomezc (dismissed):

sibille:
@AndreMaz, to get this out of preview we'd like to have an additional review from another moleculer expert, ideally from one of the creators if possible.

sibille (dismissed):

AndreMaz:
Ok @sibille I'll ask @icebob to review it

AndreMaz:
**[Update]:** This post is out of date. Please see https://github.com/microsoft/WebTemplateStudio/pull/1183#issuecomment-540496691


@sibille there is one thing that currently is not very Moleculer-ish but there is no easy way to solve it. I will explain what I mean.

At this moment the handlers are located in `pages.actions.js` and they are declared as:
https://github.com/microsoft/WebTemplateStudio/blob/516507d9b1504c43a1f73680a07948e860d03fbd/templates/Web/_composition/MoleculerJS/Page.Moleculer.MasterDetail.AddRoutes/services/pages.actions_postaction.js#L1-L13
this is not the cleanest way to declare things but it's the only way, that I've found, that works with composition templates.

The recommended way would be to declare handlers directly in the service's action block:
https://github.com/microsoft/WebTemplateStudio/blob/516507d9b1504c43a1f73680a07948e860d03fbd/templates/Web/Projects/MoleculerDefault/services/pages.service.js#L28-L32

For example something like:
```js
  /**
   * Service Actions
   * More info: https://moleculer.services/docs/0.13/actions.html
   */
  actions: {
    grid: {
      rest: "GET /grid",
      handler(ctx) {
        return sampleData.textAssets;
      }
    },
    masterDetail: {
      rest: "GET /masterdetail",
      handler(ctx) {
        return sampleData.textAssets;
      }
    }
  },
```

Now the tricky part are the commas `,` between the handlers. The way that merging algorithm works and the fact that `grid`, `masterdetail` and other handlers are located separate templates means that there is no easy way to add, when necessary, the commas.


To solve this I could put trailing commas after each handler (feature available since [ECMAScript 5](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Trailing_commas)). This way we would end up with something like this:
```js
  /**
   * Service Actions
   * More info: https://moleculer.services/docs/0.13/actions.html
   */
  actions: {
    grid: {
      rest: "GET /grid",
      handler(ctx) {
        return sampleData.textAssets;
      },
    },
    masterDetail: {
      rest: "GET /masterdetail",
      handler(ctx) {
        return sampleData.textAssets;
      },
    },
  },
```

However, in [CONTRIBUTING.md](https://github.com/microsoft/WebTemplateStudio/blob/dev/CONTRIBUTING.md#naming-and-styling-conventions) it is explicitly stated that: `All code that gets merged in must be formatted with Prettier`. The bad part is that by default Prettier has trailing commas disabled. This means that each time that I lint the code to match WebTS's style the trailing commas are deleted and, therefore, the generated templates are broken. One solution could be to create a `.prettierrc` in WebTS repo with trailing comma enabled
```js
{
  "trailingComma": "es5"
}
```
However, this would affect the entire WebTS project. Prettier would add trailing commas everywhere.

@sibille any workarounds for the trailing comma would be appreciated.

PS: I've spoken with @icebob (creator and lead dev) and he will review the templates.

icebob:
First of all, thank you @AndreMaz that you created this template for Moleculer.

I've checked the generated code and I have some suggestion:

1. All frontend builds the client files into the `server/build` folder. So we should set this path into the `assets` property in the `api.service.js` instead of `public`.
```js
    // Serve assets from "server/build" folder
    assets: {
      folder: "server/build"
    }
```

2. The Moleculer templates follow the one service - one file convention. So it would be better to merge the `pages.actions.js` file into `pages.service.js`. I mean something similar one:
```js
"use strict";

const actions = {};

actions.masterDetail = {
  rest: "GET /masterdetail",
  handler(ctx) {
    return sampleData.textAssets;
  }
};

module.exports = {
  name: "pages",

  /**
   * Service settings
   * More info: https://moleculer.services/docs/0.13/services.html#Settings
   */
  settings: {
    rest: "/"
  },
  
  /**
   * Service dependencies
   * More info: https://moleculer.services/docs/0.13/services.html#Dependencies
   */
  dependencies: [],

  /**
   * Service Actions
   * More info: https://moleculer.services/docs/0.13/actions.html
   */
  actions,

 /* ... other properties... */
}
```

3. As AndreMaz mentioned, would be good if we can use trailing comma in the last property in the `actions` object because the service file will be the same as the examples in the Moleculer documentation.

```js
"use strict";

module.exports = {
  name: "pages",

  /**
   * Service settings
   * More info: https://moleculer.services/docs/0.13/services.html#Settings
   */
  settings: {
    rest: "/"
  },
  
  /**
   * Service dependencies
   * More info: https://moleculer.services/docs/0.13/services.html#Dependencies
   */
  dependencies: [],

  /**
   * Service Actions
   * More info: https://moleculer.services/docs/0.13/actions.html
   */
  actions: {
    /**
     * Action for the Master Detail page.
     */
    masterDetail: {
      rest: "GET /masterdetail",
      handler(ctx) {
        return sampleData.textAssets;
      }
    },

    /* ... other actions ... */
  },
``` 

By the way, good job @AndreMaz!

AndreMaz:
Thanks for the input @icebob 

@sibille @dgomezc how do you want to proceed? Shall I wait for this PR to be merged and then create another PR with icebob's suggestions? Or just do the changes here?

AndreMaz:
Ok, I think I found a way to trick Prettier and move action handlers into `pages.service.js`. I'm going to refactor my templates

AndreMaz:
So, I've managed to completely get rid of `pages.actions.js`, as @icebob recommended in https://github.com/microsoft/WebTemplateStudio/pull/1183#issuecomment-540151694, without changing Prettier configs (as I initially suggested in https://github.com/microsoft/WebTemplateStudio/pull/1183#issuecomment-540026595).

Now all action handlers are placed directly in service schema:
```js
"use strict";

module.exports = {
  name: "pages",

 /* ... other configs ... */

  /**
   * Service Actions
   * More info: https://moleculer.services/docs/0.13/actions.html
   */
  actions: {
    /**
     * Action for the Master Detail page.
     */
    masterDetail: {
      rest: "GET /masterdetail",
      handler(ctx) {
        return sampleData.textAssets;
      }
    },

    /* ... other actions ... */
  },
```

AndreMaz:
Hey @sibille since in https://github.com/microsoft/WebTemplateStudio/pull/1183/commits/98f7976eed30f79125ce9d58d8f2dd5048c411bd I've refactored the templates to match @icebob's suggestions, in https://github.com/microsoft/WebTemplateStudio/pull/1183/commits/24e9b60a5f4eea25231e4d7efb63d549e9032b1c I've took Moleculer out of preview mode (as you've mentioned in https://github.com/microsoft/WebTemplateStudio/pull/1183#issuecomment-539897376).

sibille:
Thanks @AndreMaz, we'll review asap. How did you trick prettier?

AndreMaz:
Hey @sibille

So previously actions in `pages.service.js` in default template looked like this:

https://github.com/microsoft/WebTemplateStudio/blob/516507d9b1504c43a1f73680a07948e860d03fbd/templates/Web/Projects/MoleculerDefault/services/pages.service.js#L28-L32

Now, I might be wrong but I think that currently composition template only support "insert" method, meaning that there is no way to create a composition template that will change this:
```js
actions: {}, 
```
into this:
```js
 actions: {
    masterDetail: {
      rest: "GET /masterdetail",
      handler(ctx) {
        return sampleData.textAssets;
      }
    },
```
because we would need to somehow move the `},` (from `actions: {},` ) to the end.

> There is no "replace" method that will replace `},` with a composition template. Or "delete" method that will remove the `},`

I tried to solve this by changing default template into this
```js
actions: {
/* < leave some space here to insert the composition templates>*/
}, 
```
However, Prettier doesn't like empty spaces so it removes them and generates this `actions: {},`
To trick Prettier I've declared default template as:
```js
  actions: {
    // Action handlers
  },
```
With `// Action handlers` comment Prettier is happy and I'm able to easily insert action handlers.
This solved the Prettier issue in default template folder.

The other change was to declare all composition templates with the trailing comma. For example:
```js
  /**
   * Service Actions
   * More info: https://moleculer.services/docs/0.13/actions.html
   */
  actions: {
    // Action handlers
  //{[{
    grid: {
      rest: "GET /grid",
      handler(ctx) {
        return sampleData.textAssets;
      }
    }, // <- TRAILING COMMA
  //}]}
  }, 
```
In current state, this composition template is an invalid JS object meaning that Prettier won't "fix" it and remove the trailing comma.

Finally, node supports trailing commas since v8 so the generated templates are valid for all currently supported versions of node.
image

icebob:
Good job @AndreMaz

sibille (approved):

AndreMaz:
Hey @dgomezc can you please trigger the build again? I don't think that the timeout error is related with my templates

AndreMaz:
Don't know who exactly triggered the build but thanks :+1:

PR #1185, opened 2019-09-22
Updated Instructions

thekaizendiary:
Updated Instructions to install python dependencies


dgomezc:
Hi @AyushShaw, the original instruction is correct. This instruction executes the "install-requirements" script configured in package.json file that installs the flask framework.

Did you have any problems installing python dependencies?

thekaizendiary:
> Hi @AyushShaw, the original instruction is correct. This instruction executes the "install-requirements" script configured in package.json file that installs the flask framework.
> 
> Did you have any problems installing python dependencies?

I am Sorry.

Issue #1190, opened 2019-09-29
[Wizard] Cosmos Options Inconsistencies


AndreMaz:
While trying to integrate CosmosDB with Moleculer https://github.com/microsoft/WebTemplateStudio/issues/1124#issuecomment-535908869 I've noticed some inconsistencies:

Subscription is showing repeatedly "Free Trial" but I only have one (free) account.
![image](https://user-images.githubusercontent.com/9802754/65835946-9747d980-e2e4-11e9-9611-5b6a05dac3cf.png)

In API section I'm getting two options (Mongo and SQL). I'm currently working only on Mongo integration so SQL should not appear in options. I think some kind of check and validation should be done here.
![image](https://user-images.githubusercontent.com/9802754/65835967-d24a0d00-e2e4-11e9-95f4-e02a2a83c547.png)

dgomezc:
Hi @AndreMaz,

Web Template Studio has an internal dependency on the Azure Account extension, that provides Azure Subscription data. Can you execute "Azure: Select Subscriptions" command in VSCode and confirm that show the same subscriptions?

In API section, we can select MongoDB or SQL because this options are selectables when create a MongoDB services in Azure Portal, and some users prefer SQL instead of MongoDB.

AndreMaz:
Hi @dgomezc 

This is what I get when I run "Azure: Select Subscriptions"
![image](https://user-images.githubusercontent.com/9802754/65893909-ba37c380-e3a0-11e9-9154-18e0df907211.png)

> some users prefer SQL instead of MongoDB.

I absolutely agree with you. But what I'm trying to say is that, at this moment, Moleculer does not support Cosmos SQL API. There is no composition template for that. So the presence of SQL gives a wrong impression to the end user.

Same might happen with other frameworks that will be added in the future.  For example, imagine that after a while you decide to add support for Cassandra (one of several APIs supported by Azure). What would happen then? Will you update all back-end frameworks at the same time? If no, then the same situation would happen: the combination "framework X + Cassandra" would appear as a valid option to the end user. Even if framework X does not support Cassandra.

So I think that CosmosDB API selection field should only show a set of valid options, i.e.,  APIs that the selected framework (Step 2) has support.

dgomezc:
Hi @AndreMaz, 

I could not reproduce the problem, but when reviewing the code, I have seen that it can be related to azure sessions. To fix this, I will verify that you cannot add a subscription to the combobox if one with the same Id already exists.

[...]

PR #1196, opened 2019-10-03
Filter repeated Azure subscriptions

dgomezc:
x-ref: #1190


dgomezc:
Hi @AndreMaz,  Can you test with this PR to verify that there are no repeated subscriptions?

AndreMaz:
Hey @dgomezc. Good job :+1: repeated subscriptions are gone

sibille (dismissed):

src/extension/src/azure/azure-auth/azureAuth.ts
```diff
@@ -125,14 +125,20 @@ export abstract class AzureAuth {
...
-        ...subscriptions.map(subscription => ({
-          label: subscription.displayName || "",
-          subscriptionId: subscription.subscriptionId || "",
-          session,
-          subscription
-        }))
-      );
+
+      let filteredSubscriptions = subscriptions.reduce((items, subscription) => {
+        if (subscription.subscriptionId && !(subscriptionItems.some(item => item.subscriptionId === subscription.subscriptionId))) {
```
jcoc611-microsoft: NIT: This seems a bit inefficient (`O(n^2)` where we could do `O(n)` by filtering the list in one pass). I understand that most people will have very few subscriptions though. Also, where are the duplicate subscriptions coming from? Is there a way to query for unique list of subscriptions instead?

dgomezc: Hi @jcoc611-microsoft,

We  could not reproduce the issue, but we think it is related with sessions (posibility two opened session with same subscriptionId?). 

the subscriptions filter could be:

```javascript
subscriptionItems.push(
        ...subscriptions
        .filter(subscription => 
          subscription.subscriptionId && 
          !(subscriptionItems.some(item => item.subscriptionId === subscription.subscriptionId)))
        .map(subscription => ({
          label: subscription.displayName || "",
          subscriptionId: subscription.subscriptionId || "",
          session,
          subscription
        }))
      );
```
If you agree we will update the code?

jcoc611-microsoft: Makes sense.

What do you think of the following approach?
```TypeScript
let subscriptionItems: SubscriptionItem[] = [];
let subscriptionIds: Map<string, boolean> = new Map();
for (const session of this.api.sessions) {
    // ..
    for (const subscription of subscriptions) {
        if ( subscriptionIds.has( subscription.subscriptionId ) ) {
            continue;
        }
        subscriptionIds.set( subscription.subscriptionId, true );
        subscriptionItems.push( ... )
    }
}
```

dgomezc: In your example the subscriptionId check is missing. 

How about checking against the subscriptionItems array instead of subscriptionIds map?

I also remove the continue statement.

```javascript
let subscriptionItems: SubscriptionItem[] = [];
for (const session of this.api.sessions) {
    // ..
    for (const subscription of subscriptions) {
        if (subscription.subscriptionId && !subscriptionItems.some(item => item.subscriptionId === subscription.subscriptionId )) {
            subscriptionItems.push( ... )
        }        
    }
}
```
What do you think about this approach?

jcoc611-microsoft: The issue I have with this approach is that `subscriptionItems.some` loops through the subscription array every iteration of the loop. If performance is not a concern then it is fine. I'm not sure I understand why it is functionally different from the approach I suggested.

sibille: Got it! We're going with your approach, only included the subscriptionId check and inverted the continue, as this seams easier to understand with two conditions.

jcoc611-microsoft (approved):

sibille (approved):

Issue #1124, opened 2019-08-21
Add MoleculerJS


AndreMaz:
[Moleculer](https://moleculer.services/) is a progressive microservices framework for Node.js. It is highly customizable and has a lot of [cool features](https://moleculer.services/docs/0.13/index.html#Features). It also has a [project template](https://github.com/moleculerjs/moleculer-template-project#moleculer-template-project) that can be modified/adapted, and be added to the WebTemplateStudio.

Tanya0609:
Hi @AndreMaz, thank you so much for your suggestion! It seems a great framework to consider, we will add it into backlog and investigate when we have dev capability! 

Also, if you are interested in contributing, please feel free to look at the [contribution guidelines](https://github.com/microsoft/WebTemplateStudio/blob/dev/CONTRIBUTING.md).

AndreMaz:
Hey @Tanya0609. Thanks for your response.

I can definitely help with the template but I need some guidance. Where should I put the metadata (description, logo, and etc.) about Moleculer? I mean, how can I add a new framework to the list back-end frameworks? The [Add Backend Framework](https://github.com/microsoft/WebTemplateStudio/wiki/Templates#add-backend-framework) Wiki has no mention about it.

AndreMaz:
So I've managed to get things working
Wizard: 
![image](https://user-images.githubusercontent.com/9802754/64132980-50d98a80-cdcb-11e9-9c74-fa8d0e55f2c6.png)
Moleculer+Vue
![FM2eNnx8T4](https://user-images.githubusercontent.com/9802754/64133011-7e263880-cdcb-11e9-9fff-9ea4ff764ad0.gif)

The only thing that is missing is the integration with Azure Cosmos.

Is there any interest in having this added to WebTemplateStudio? If so, then I can make a PR.

smmatte:
Hi @AndreMaz this looks great, if you make a PR, then we would look at it. Thank you!

AndreMaz:
@smmatte PR https://github.com/microsoft/WebTemplateStudio/pull/1173 got 3 approvals so I'm assuming that it eventually is going to be merged. Following this line of thought I want to discuss the contents of next PR. 

I was thinking about adding unit tests (Moleculer plays well with Jest) and Docker files. What do you think? 

----

[...]

PR #1208, opened 2019-10-23
[moleculer] Refactor templates

AndreMaz:
This PR contains refactored Moleculer templates in order to allow Azure App Service deployments.
More info here: https://github.com/microsoft/WebTemplateStudio/issues/1124#issuecomment-544590572 , https://github.com/microsoft/WebTemplateStudio/issues/1124#issuecomment-544656829

I've refactored the templates to match `Node/Express` file structure. 

- App logic is located in `server` folder
- Two `package.json` files. 1 for dev and 1 for production

image

@sibille @dgomezc can you test it and see if it works properly?


sibille (commented):
Great job @AndreMaz, I got this working on App Service with two adjustments as only the server folder should be deployed to App Service. 
Actually a settings.json file should be created with the url and folder to deploy like:
```
{
    "appService.defaultWebAppToDeploy": "[yourAppServiceURL]",
    "appService.deploySubpath": "server"
}
```
This is on our side and I'll look at why this isn't generated.

Adoptions I made to deploy "server" to Azure App Service
- In the server package.json start statement should be `"start": "moleculer-runner services"`
- Data folder should be a subfolder of "server" to get deployed to

AndreMaz:
Thanks for feedback @sibille 😃 I will update my PR later today

AndreMaz:
Should be fixed now @sibille

AndreMaz:
Please don't merge this yet. I forgot to move `moleculer.config.js` into `server` folder. 
It will work even without this file because if config file is missing moleculer assumes default configs.  
However, the absence of this file might seem confusing to users.

I will move it right now

AndreMaz:
@dgomezc @sibille quick question: during Azure App Service deployment, what happens with environment variables (`.env` file) with DB connection info? Is it copied into `server` folder and then moved into App Service?

AndreMaz:
Looking at `node/express` example it seems that `.env` file is present in folder copied into App Service. I'll do the same for moleculer

```js
const CosmosClient = require("@azure/cosmos").CosmosClient;
require("dotenv").config();

// Connects the back-end to the Cosmos Core SQL Database (https://docs.microsoft.com/en-us/azure/cosmos-db/create-sql-api-nodejs)
module.exports = class SQLClient {
  constructor(databaseId, containerId) {
    this.client = new CosmosClient({
      endpoint: process.env.COSMOSDB_URI,
      auth: {
        masterKey: process.env.COSMOSDB_PRIMARY_KEY
      }
    });
```

AndreMaz:
All good on my side. Please review it and merge if everything is Ok.

sibille:
.env file does not need to be in server folder, the connection string is set on the App Service after cosmosdb creation by the extension. Did you change it?

AndreMaz:
Yes, right now it looks like `"start": "moleculer-runner --env services"`. With `--env` param `moleculer-runner` will try to read `.env` in `.`. If there is no file it will ignore and move forward.

I can remove it if you want

sibille:
I think it's fine, but let me double check

sibille:
@AndreMaz, for Azure App Service deployment "--env" is not needed, application settings are converted into environment variables there. Nevertheless I'd go back to what you had before to allow local execution from the server package.json: 
`moleculer-runner --envfile ../.env services` .

We saw the config file is not specified in the server package.json, is it grabbed automatically as it is in the same folder?

AndreMaz:
Hey @sibille. I've reverted back to  `moleculer-runner --envfile ../.env services` 

> We saw the config file is not specified in the server package.json, is it grabbed automatically as it is in the same folder?

With `--env` param `moleculer-runner` will look for `.env` in the same folder. For custom file/location `--envfile` should be used. So in top-level `./package.json` I'm using `--env` and in `./server/package` the `--envfile`

---

Initially I just tried to mimic the `node/express` behavior that uses `require('dotenv').config()` in `./server/slq/sqlClient.js`. 
image

Because of this line I've assumed that `.env` file is copied in `server` folder and then moved into Azure App.

> for Azure App Service deployment "--env" is not needed, application settings are converted into environment variables there

Also, for `node/express`, the `require('dotenv').config()` won't work if I were to run locally the `npm start` from `./server/package.json` as there is not `.env` file in `server` folder. I think it's a bug.

dgomezc (approved):

PR #1221, opened 2019-10-31
Feedback on generated React code

### LINKED ISSUE #1067: Feedback on generated React code
The following feedback are from one of the React experienced users:

- [x] class property init (rather than instructor)
- [x] the way to bind the class functions 
- [x] Naming convention 
i.e. Grid.state.WarningMessageOpen normally camel case instead of pascal for variables
i.e. folder name should have underscore i.e. 'Master_Detail'
- [x] for constants, styles and images, don't recommend relative paths, should use absolute paths
- [x] WarningMessage/index.jsx the function name should be WarningMessage instead of 'index', this is sometimes useful for debugging
- [x] putting classnames (inline) is a bad practice, it will create new css class every re-render
- [x] registering service worker by default should be ok but you should explain to people that the website will only update to latest version on 2nd visit (instead of first) if it is already cached

mcmxcdev:
> The following feedback is from one of the React experienced users:
> * [x]  class property init (rather than instructor)
> * [x]  the way to bind the class functions
> Naming convention
> * [x]  i.e. Grid.state.WarningMessageOpen normally camel case instead of pascal for variables
> * [ ]  i.e. folder name should have underscore i.e. 'Master_Detail'
> * [ ]  for constants, styles and images, don't recommend relative paths, should use absolute paths
> * [x]  WarningMessage/index.jsx the function name should be WarningMessage instead of 'index', this is sometimes useful for debugging
> * [x]  putting classnames (inline) is a bad practice, it will create new css class every re-render
> * [ ]  registering service worker by default should be ok but you should explain to people that the website will only update to latest version on 2nd visit (instead of first) if it is already cached

I didnt rename the folders, because:
* it would need to be done for all the generated code for all 3 frameworks to be consistent, not only Angular
* it would potentially screw up the git commit history and lead to a lot of changed files because the whole sub file tree would be counted as newly added.

I didnt use absolute paths, because:
* I never saw this being actually used in any frontend framework
*  it seems impractical because as soon as any path segment would change, the import would be wrong
* auto import extensions (vscode, intellij) usually resolve imports with relative paths by default

Regarding service worker registering I would need some information about where to put it, with what content and in what form (could be an alert message, could be a `console.warn`, etc)

Closes #1067

### DIFF
--- a/src/client/src/containers/SelectFrontEndFramework/index.tsx
+++ b/src/client/src/containers/SelectFrontEndFramework/index.tsx
@@ -60,7 +60,7 @@ const messages = defineMessages({
 });
 
 class SelectFrontEndFramework extends React.Component<Props> {
-  public handleFrameworkChange(option: ISelected) {
+  public handleFrameworkChange = (option: ISelected) => {
     const {
       vscode,
       selectedFrontendFramework,
@@ -136,7 +136,7 @@ class SelectFrontEndFramework extends React.Component<Props> {
       <div>
         {this.props.options.length > 0 && (
           <SelectOption
-            selectCard={this.handleFrameworkChange.bind(this)}
+            selectCard={this.handleFrameworkChange}
             multiSelect={false}
             title={intl.formatMessage(messages.selectFrontendFramework)}
             isFrameworkSelection={true}
--- a/templates/Web/Pages/React.Grid/src/components/ReactGrid/GridComponent.jsx
+++ b/templates/Web/Pages/React.Grid/src/components/ReactGrid/GridComponent.jsx
@@ -1,7 +1,6 @@
 ﻿import React from "react";
 
-export default function GridComponent(props) {
-  const { image, header, description } = props;
+export default function GridComponent({ image, header, description }) {
   return (
     <div className="col-md-4 col-sm-12 p-5">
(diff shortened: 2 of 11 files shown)


jcoc611-microsoft (dismissed):
Great job! Thank you!

sibille:
> I didnt rename the folders, because:
> it would need to be done for all the generated code for all 3 frameworks to be consistent, not only Angular
> it would potentially screw up the git commit history and lead to a lot of changed files because the whole sub file tree would be counted as newly added.

We do casing variations for Angular already, but could not find any official guidelines for React naming and casing. So I think for React we are fine with generating the name the way the user introduces it in the wizard. 

> I didnt use absolute paths, because:
> I never saw this being actually used in any frontend framework
> it seems impractical because as soon as any path segment would change, the import would be wrong
> auto import extensions (vscode, intellij) usually resolve imports with relative paths by default

Agree! 

> Regarding service worker registering I would need some information about where to put it, with what content and in what form (could be an alert message, could be a console.warn, etc)

We'd suggest to add a comment to the index.js file the way create-react-app does. For example:
```javascript
// Learn more about service workers in React: https://create-react-app.dev/docs/making-a-progressive-web-app
registerServiceWorker();
```
Do you think this is sufficient or do we need more info?

dgomezc (approved):

Issue #359, opened 2019-04-08
Adding and removing pages does not reset the page name counter.


JimmyH936:
This is the case for all pages.

Steps to reproduce:

1. Launch Wizard
2. Go to pages
3. Delete the default added Blank Page.
4. re-add the blank page

Under project details, it will appear as (Blank2) instead of just (Blank).

Tanya0609:
This is related to re-design the paging name logic.

sibille:
Fixed as part of https://github.com/microsoft/WebTemplateStudio/pull/1309

Verified in dev-nightly:
Templates version: 0.1.20028.1
Wizard version: 0.1.2002801

PR #1249, opened 2019-11-20
Reset the page name counter

javieraparisivaldes:
## PR checklist

- Quick summary of changes
reset the page name counter, add new funcionality to suggest new name looking since the first array position (to recover, ex: 'blank2' if your last title is 'blank4' and the blank2 was deleted), create unit test and refactor function names and change es5 to es6 code 
#359


sibille (changes requested):
Need to fix right side bar behavior

Issue #1261, opened 2019-12-02
Update libraries to latest version in React templates


sibille:
We should use fixed library versions in dependencies to avoid issues like https://github.com/microsoft/WebTemplateStudio/issues/1282

dgomezc:
Verified in dev-nightly ;
Templates version: 0.1.20052.2
Wizard version: 0.1.2005202

PR #1269, opened 2019-12-05
Update react libs

javieraparisivaldes:
## PR checklist

- Quick summary of changes
Update react libs
#1261 
- Anything that requires particular review or attention?
You need test React, Angular and Vue proyects


src/extension/logs/wts.log..2019-12-04
```diff
@@ -0,0 +1,2 @@
+[2019-12-05T16:24:37.359] [INFO] default - [ EXTENSION ]  Launched
```
jcoc611-microsoft: I think you added this one by mistake?

sibille: yes, we should add the logs folder to .gitignore

Issue #1635, opened 2020-10-20
Dev-nightly: Increased font size on browse button after navigating to step 5 in wizard


sibille:
**Describe the bug**
Increased font size on browse button after navigating to step 5 in wizard

**To Reproduce**
1. Start Web Template Studio selecting "Web Template Studio: Launch" from VS Command pallete
2. Click on Step 5 "Summary"
3. Click on Step 1 "New project"
4. Observe increased font size on button "Browse"

**Expected behavior**
Font size should not be increased
![image](https://user-images.githubusercontent.com/24698018/96601339-32499200-12f2-11eb-81b2-c334b2159680.png)

**Screenshots**
![image](https://user-images.githubusercontent.com/24698018/96601213-134b0000-12f2-11eb-93c5-a6556673b479.png)

sibille:
Verified in nightly version: 
Templates version: 0.3.20297.1
Wizard version: 0.3.2029701

Issue #1376, opened 2020-03-09
Images should be read from templates folder


sibille:
We should read the images from the templates folder (through the extension) instead of having them duplicated on the client.

For example, the node image should come from https://github.com/microsoft/WebTemplateStudio/blob/dev/templates/Web/_catalog/backendframeworks/Node.svg

prupipho:
Updated images that now comes from the templates folder:
- frontend & backend framework
- pages
- azure services

prupipho:
Checked on nightly version:

Templates version: 0.3.20287.1
Wizard version: 0.3.2028701

PR #1640, opened 2020-10-21
default image when missing in template

prupipho:
## PR checklist

- Quick summary of changes
default image 
fix resize of button text
same images for mock in client
#1635 , #1376


prupipho:
Seem to be an existing unknown issue #1641 

Wrong icon shows in the rightsidebar  => this seems to be an existing issue. As we can see on the second screenshot, if we also change the name of the page on the template, it updates on the main page, but not on the rightsidebar.

image

image

sibille (approved):

Issue #1595, opened 2020-09-10
Update npm/nuget packages to latest version in generated app


sibille:
Update npm packages to latest version in generated app

prupipho:
Verified in nightly build:

Templates version: 0.3.20318.2
Wizard version: 0.3.2031802

Issue #1494, opened 2020-06-22
Change generated web applications folder structure 


sibille:
We're looking into changing the folder structure of the web applications we generate to use separate folders for frontend and backend.
 
We've setup a sample project at https://github.com/dgomezc/StructureProposalApp and would love to get your feedback on this, to ensure what we generate works for everybody.
 
The idea is that each project lives in its own folder with it's own package.json or scripts. For deployment we'll create a third folder with the server and compiled frontend. We're also adding VSCode Tasks to install dependencies and start client and server concurrently or separately for development.
 
Questions:
	• Does this new structure match your needs or do you see any issues? 
	• Is there anything else we should add? (f. ex scripts)

For more info on the motivation for this propsal see #1209 and #1189

AndreMaz:
Are you planning to add `e2e` tests? If so, where would you place them?

sibille:
We're not actively working on adding e2e tests, but it's a good idea, will open an issue to add it in. We think e2e tests should be located in the front-end folder, that's where the angular-cli puts them and we think it makes sense as they test the frontend (integrated with the backend).

marko-hologram:
I have few thoughts, I'll try to split them in categories like folders or files, depending on which part of the template I'm giving feedback for. I do like the idea of separate backend and frontend folders and generating build files into a third folder, but I do have some thoughts overall. I'll also address this from React perspective since the repo that's linked is a React project.

## Folders

Maybe renaming `/backEnd` to `/server` and `/frontEnd` to `/client`? I haven't actually seen that much camel cased folder names "out in the wild". And it also might look weird considering the fact that "arm-templates" is dash separated (kebab-case). Or if you really want to leave it as backend and frontend name in particular, I would suggest against camelCase possibly. (`/frontend`, `/backend`).

One thing also I've seen that one folder is named "Master_Detail" in "frontEnd/src/components". I also find this highly unusual since other folders in components folder are named with PascalCase (NavBar, WarningMessage). Usually React convention is PascalCase for component names and folder names as far as I've seen. I see that in a newly generated project that component is called MasterDetail, while in this example repo it's called Master_Detail. Also, I'm not sure if word "master" currently clashes with the fact you are trying to rename master branch to something else (https://github.com/microsoft/WebTemplateStudio/issues/1475), so maybe even renaming "MasterDetail" component is something to keep in mind.

[...]

PR #1651, opened 2020-10-28
1595 update packages moleculer 

prupipho:
## PR checklist

- Quick summary of changes
update packages moleculer
#1595, #1656

### DIFF
--- a/templates/Web/Projects/MoleculerDefault/.vscode/tasks_postaction.json
+++ b/templates/Web/Projects/MoleculerDefault/.vscode/tasks_postaction.json
@@ -20,7 +20,7 @@
       "label": "Start backend",
       //{[{
       "type": "npm",
-      "script": "start",
+      "script": "start-dev",
       "path": "backend/"
       //}]}
     },
--- a/templates/Web/Projects/MoleculerDefault/README_postaction.md
+++ b/templates/Web/Projects/MoleculerDefault/README_postaction.md
@@ -23,7 +23,7 @@ The most important scripts in the `package.json` are:
 To start the backend application manually:
   1. Open a terminal and navigate to the `backend` folder path.
   2. Use `yarn install` or `npm install` to install backend dependencies.
-  3. Use `yarn start` or `npm start` to start backend app in development.
+  3. Use `yarn start-dev` or `npm start-dev` to start backend app in development.
 //}]}
 
 ## Additional Documentation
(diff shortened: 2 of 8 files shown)


sibille:
Looks good to me!
@AndreMaz can you confirm moleculer setup and versions are correct after the project reorg from #1494 and the updates from this PR?

AndreMaz:
Hey @sibille 

Regarding the versions, there should be no problem as there were no breaking changes in the latest releases.

Now regarding the project reorg, I've done a quick test (without CosmosDB) and it's looking good. I only see a tiny issue in the `package.json`

```json
"scripts": {
    "start-dev": "moleculer-runner services",
    "start": "moleculer-runner --repl --hot --config ./moleculer.config.js ./services",
  },
```
I'm assuming that `start-dev` is for development and `start` is for production. Right? If so then I would say that they should be swapped.

The `--repl` flag is used to enable the REPL terminal and the `--hot` is for hot reloading the services. Also, the `--config ./moleculer.config.js` is not necessary because `moleculer-runner` looks for the `moleculer.config.js` by default. The `--config` should be used when the configs are in a different file or dir. The `services` option indicates the dir where the services are located. So `moleculer-runner` will look at that dir and load all the files that have the following naming scheme: `<something>.service.js`

Having said that, I think that this is how things should look
```json
  "scripts": {
    "start-dev": "moleculer-runner --repl --hot services",
    "start": "moleculer-runner services"
  },
```

sibille:
Thanks @AndreMaz. you're right! 
start-dev is for starting the backend in development, so according to the parameters you explained switching makes sense. I've adapted the task.json and readme.md in this PR too to fix this

AndreMaz:
@sibille I didn't test the cosmosDB template but as far as I remember there was an `.env` file with password for the DB.
For loading the env vars you have 2 options:
```bash
# Load the default .env file from current directory
moleculer-runner --env services

# Load the specified .my-env file
moleculer-runner --envfile .my-env services
```

sibille:
That's correct. Thanks for the info! I'll test and adjust the compositions for cosmos db.

sibille:
Done, with Cosmos DB configuration is:
```
    "start-dev": "moleculer-runner --repl --hot --env services",
    "start": "moleculer-runner --envfile ./.env services",
```

sibille (approved):

PR #1653, opened 2020-10-28
1595 update packages vue

prupipho:
## PR checklist

- Quick summary of changes
update packages vue
#1595

### DIFF
--- a/templates/Web/Pages/Vue.List/frontend/src/views/VueList.vue
+++ b/templates/Web/Pages/Vue.List/frontend/src/views/VueList.vue
@@ -5,18 +5,18 @@
     </div>
     <div class="row">
       <div class="col-12 p-0">
-        <ListForm v-model="textField" @addListItem="addListItem" />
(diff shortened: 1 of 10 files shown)


templates/Web/Pages/Vue.List/frontend/src/views/VueList.vue
```diff
@@ -5,18 +5,18 @@
     </div>
     <div class="row">
       <div class="col-12 p-0">
-        <ListForm v-model="textField" @addListItem="addListItem" />
+        <ListForm v-model="textField" @add-list-item="addListItem" />
```
prupipho, line 8: @sibille @dgomezc  not 100% sure but I think this needs to match with the $emit( ) one. I´d suggest changing the function name "addListItem" to something like in the onWarningClose, so it is clear that they are two different things.

Example:
``` add-list-item=handleAddListItem```

prupipho:
I think some changes won´t be required

